PR #698, opened 2023-06-06
Staking dashboard UI preparation

ERussel:
- add model builder to gather data from different source and compose single list of models for presenter
- add view model factory to transform dashboard item model to view model


novawallet/Modules/Staking/Dashboard/Model/StakingDashboardModel.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    var hasStaking: Bool {
+        guard let dashboardItem = dashboardItem else {
+            return false
+        }
+
+        return dashboardItem.stake != nil
+    }
+
+    func balanceValue() -> Decimal {
```
lynx56, line 21: looks like a property

ERussel: I think we should mark as property the logic that for constant time, otherwise better to use functions

PR #702, opened 2023-06-09
Web3 Names: support KiltTransferAssetRecipientV2

lynx56:


ERussel (commented):
Looks good. Let't test and promote

novawallet/Common/Web3Names/JsonCanonicalizer.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    private func canonicalize(_ object: Any) -> String {
+        var result = ""
+
+        if let string = object as? String {
+            result += "\"\(canonicalize(string: string))\""
+        } else if let array = object as? [Any] {
+            result += "["
+            var next = false
+            array.forEach { element in
```
ERussel: I think the cleaner way would be: ```array.map { ... }.joined(",")```. The same can be applied to dictionary case.

novawallet/Common/Web3Names/JsonCanonicalizer.swift
```diff
@@ -0,0 +1,101 @@
...
+        } else if let dictionary = object as? [String: Any] {
+            result += "{"
+            var next = false
+            dictionary.keys.sorted { (str1, str2) -> Bool in
+                let utfView1 = Array(str1.utf16)
+                let utfView2 = Array(str2.utf16)
+
+                for index in 0 ..< max(utfView1.count, utfView2.count) {
+                    if let char1 = utfView1[safe: index], let char2 = utfView2[safe: index] {
+                        if char1 == char2 {
```
ERussel, line 43: I think more cleaner way is to just check for inequality. In this case we don't need to use ```continue```.

ERussel (approved):

ERussel (commented):
Let's address the comments before merge

novawallet/Modules/Transfer/TransferSetup/TransferSetupViewFactory.swift
```diff
@@ -112,7 +112,9 @@ struct TransferSetupViewFactory {
         let operationQueue = OperationManagerFacade.sharedDefaultQueue
         let web3NamesOperationFactory = KiltWeb3NamesOperationFactory(operationQueue: operationQueue)
 
-        let recipientRepository = KiltTransferAssetRecipientRepository(integrityVerifier: Web3NameIntegrityVerifier())
+        let recipientRepositoryFactory = Web3TransferRecipientRepositoryFactory(integrityVerifierFactory: Web3TransferRecipientIntegrityVerifierFactory())
+
+        KiltTransferAssetRecipient.Version1.Repository(integrityVerifier: Web3NameIntegrityVerifier())
```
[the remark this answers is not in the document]
lynx56: why?

PR #712, opened 2023-06-17
Fix crosschain UI

ERussel:
- increase tap area for destination chain selector
- enable scroll for long destination chains list
- make w3n names order fixed
- fix race condition image loading

### DIFF
--- a/novawallet/Common/ViewModel/RemoteImageViewModel.swift
+++ b/novawallet/Common/ViewModel/RemoteImageViewModel.swift
@@ -49,7 +49,10 @@ extension RemoteImageViewModel: ImageViewModelProtocol {
     }
 
     func cancel(on imageView: UIImageView) {
-        imageView.kf.cancelDownloadTask()
+        imageView.kf.cancelDownloadTask() // cancel any dowload task
+
+        let url: URL? = nil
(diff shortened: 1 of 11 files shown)


novawallet/Common/ViewModel/RemoteImageViewModel.swift
```diff
@@ -49,7 +49,10 @@ extension RemoteImageViewModel: ImageViewModelProtocol {
     }
 
     func cancel(on imageView: UIImageView) {
-        imageView.kf.cancelDownloadTask()
+        imageView.kf.cancelDownloadTask() // cancel any dowload task
+
+        let url: URL? = nil
+        imageView.kf.setImage(with: url) // cancel any cache retrieval task
```
lynx56, line 55: I think it can be shorter: imageView.kf.setImage(with: nil)

lynx56 (approved):

novawallet/Common/ViewModel/RemoteImageViewModel.swift
ERussel: I tried by there are several methods with the same names for the parameter by different types. So we need to specify type to avoid ambiguity.

PR #726, opened 2023-06-25
Fix scientific notation

ERussel:
- support scientific notation for BigUInt to properly parse Subquery responses
- fix case when adding wallet but staking screen have not been opened yet (no event can be received)

### DIFF
--- a/novawallet/Common/Network/Subquery/SubqueryMultistakingOperationFactory.swift
+++ b/novawallet/Common/Network/Subquery/SubqueryMultistakingOperationFactory.swift
@@ -100,7 +100,7 @@ extension SubqueryMultistakingOperationFactory: MultistakingOffchainOperationFac
                         return
                     }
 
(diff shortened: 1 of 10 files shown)


novawallet/Common/Network/Subquery/SubqueryMultistakingOperationFactory.swift
```diff
@@ -100,7 +100,7 @@ extension SubqueryMultistakingOperationFactory: MultistakingOffchainOperationFac
                         return
                     }
 
-                    $0[.init(networkId: networkId, stakingType: stakingType)] = $1.sum.amount
+                    $0[.init(networkId: networkId, stakingType: stakingType)] = BigUInt(scientific: $1.sum.amount)
```
lynx56, line 103: is it always in scientific notation?

lynx56 (approved):

PR #777, opened 2023-08-08
Recommendation improvements

ERussel:
- take pool status and whether there is place to join while recommending
- take into account whether a user can stake with nomination pools on recommending

### DIFF
--- a/novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
+++ b/novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
@@ -1,5 +1,6 @@
 import Foundation
 import RobinHood
+import SubstrateSdk
 
 protocol NPoolsLocalSubscriptionFactoryProtocol {
     func getMinJoinBondProvider(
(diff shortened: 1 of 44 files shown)


novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
```diff
@@ -105,4 +121,64 @@ extension NPoolsLocalSubscriptionFactory: NPoolsLocalSubscriptionFactoryProtocol
...
+            missingEntryStrategy: missingEntryStrategy
+        )
+
+        return try getDataProvider(
+            for: localKey,
+            chainId: chainId,
+            storageCodingPath: codingPath,
+            fallback: fallback
+        )
+    }
```
lynx56, line 183: I think we can use one common method with codingPath input

novawallet/Modules/Staking/StakingSetupAmount/Model/StakingTypeBalanceFactory.swift
```diff
@@ -0,0 +1,50 @@
...
+    init(stakingType: StakingType?) {
+        self.stakingType = stakingType
+    }
+
+    var stakingTypeAllowsLocks: Bool {
+        switch stakingType {
+        case .relaychain, .auraRelaychain, .azero, .none, .parachain, .turing:
+            return true
+        case .nominationPools, .unsupported:
+            return true
```
lynx56, line 23: is it correct?

lynx56 (approved):

novawallet/Modules/Staking/StakingSetupAmount/Model/StakingTypeBalanceFactory.swift
ERussel: No, but I have fixed this in a separate PR

PR #779, opened 2023-08-09
Move changelog to the main repo

stepanLav:


ERussel:
We have decided not to merge this logic with the client code for safety

PR #792, opened 2023-08-18
NominationPools search 

lynx56:


novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchInteractor.swift
```diff
@@ -0,0 +1,123 @@
...
+        clear(cancellable: &poolsCancellable)
+
+        let maxApyWrapper = rewardEngineOperationFactory.createEngineWrapper(
+            for: eraNominationPoolsService,
+            validatorRewardService: validatorRewardService,
+            connection: connection,
+            runtimeService: runtimeService
+        )
+
+        let poolStatsWrapper = poolsOperationFactory.createAllPoolsInfoWrapper(
```
ERussel, line 63: I would recommend to union this two wrapper into one and then save as cancellable. Otherwise we only can cancel operations from the second wrapper

novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchPresenter.swift
```diff
@@ -0,0 +1,154 @@
...
+    func setup() {
+        interactor.setup()
+        view?.didReceivePools(state: .loaded(viewModel: []))
+    }
+
+    func search(for text: String) {
+        guard let searchOperationClosure = searchOperationClosure else {
+            return
+        }
+        if text.isEmpty {
```
ERussel, line 67: Looks like logic for the interactor

novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchViewFactory.swift
```diff
@@ -0,0 +1,72 @@
...
+        let chainRegistry = ChainRegistryFacade.sharedRegistry
+
+        guard
+            let runtimeService = chainRegistry.getRuntimeProvider(for: chainId),
+            let connection = chainRegistry.getConnection(for: chainId),
+            let activePoolService = state.activePoolsService else {
+            return nil
+        }
+
+        let queue = OperationQueue()
```
ERussel, line 54: Let's reuse OperationManagerFacade.sharedDefaultQueue

ERussel (changes requested):
- typing fast lead to the invalid state

invalid-searchstate

- check localisation and style (icon and text should have same colours) for empty state

localization-and-style

- let's display empty state in the center otherwise it looks weird

emty-state

novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchPresenter.swift
```diff
@@ -0,0 +1,154 @@
...
+    private var searchOperationClosure: NominationPoolSearchOperationClosure?
+
+    init(
+        interactor: NominationPoolSearchInteractorInputProtocol,
+        wireframe: NominationPoolSearchWireframeProtocol,
+        viewModelFactory: StakingSelectPoolViewModelFactoryProtocol,
+        chainAsset: ChainAsset,
+        searchOperationFactory: NominationPoolSearchOperationFactoryProtocol,
+        delegate: StakingSelectPoolDelegate,
+        operationQueue: OperationQueue
```
ERussel, line 27: Missing localisation manager

novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchPresenter.swift
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    func selectPool(poolId: NominationPools.PoolId) {
+        guard let pool = poolStats.first(where: { $0.poolId == poolId }) else {
+            return
+        }
+
+        if pool.maxApy == nil {
+            showUnsupportedPoolStateAlert()
```
ERussel, line 96: We should allow to join pool with not elected validators

novawallet/Modules/Staking/NominationPoolSearch/NominationPoolSearchPresenter.swift
lynx56: why? We search in viewModels

ERussel: It is not trivial and performs asyncroniously

ERussel (approved):

PR #799, opened 2023-08-24
Select staking validators

lynx56:


novawallet/Modules/Staking/StakingType/StakingTypePresenter.swift
```diff
@@ -19,6 +19,7 @@ final class StakingTypePresenter {
     private var method: StakingSelectionMethod?
     private var selection: StakingTypeSelection
     private var hasChanges: Bool = false
+    private var recommendedValidators: PreparedValidators?
```
ERussel, line 22: Not sure if we are really need this. As I see we can just pass information whether the selected validation are actually recommended from the CustomValidators module. What do you think?

novawallet/Modules/Staking/StakingType/StakingTypeInteractor.swift
```diff
@@ -66,9 +67,15 @@ extension StakingTypeInteractor: StakingTypeInteractorInputProtocol {
             $0.delegate = self
             $0.start()
         }
+
+        if case .direct = stakingSelectionMethod.selectedStakingOption {
```
ERussel, line 71: Don't really want to have staking specific logic in the universal class. That's why we have made recommendation logic so complicated in terms of architecture. Can you, please, highlight a use case for that?

novawallet/Modules/Staking/StakingSelectPool/StakingSelectPoolViewFactory.swift
```diff
@@ -6,6 +6,10 @@ protocol StakingSelectPoolDelegate: AnyObject {
     func changePoolSelection(selectedPool: NominationPools.SelectedPool, isRecommended: Bool)
 }
 
+protocol StakingSelectValidatorsDelegate: AnyObject {
```
ERussel, line 9: Didn't get what is this guy doing in the file for nomination pools

ERussel (changes requested):
After taping on "Show selected" button we should go to the selected validators screen and only after taping continue there we are returning to the Staking Type screen.

selectedValidatorsMissing

novawallet/Modules/Staking/StakingType/StakingTypePresenter.swift
```diff
@@ -324,22 +317,22 @@ extension StakingTypePresenter: StakingSelectPoolDelegate {
...
 extension StakingTypePresenter: StakingSelectValidatorsDelegate {
     func changeValidatorsSelection(validatorList: [SelectedValidatorInfo], maxTargets: Int) {
-        guard let recommendedValidators = recommendedValidators,
-              case let .direct(validators) = method?.selectedStakingOption,
+        guard case let .direct(validators) = method?.selectedStakingOption,
               let restrictions = method?.restrictions else {
             return
         }
         let selectedAddresses = validatorList.map(\.address)
         let selectedValidators = validators.electedValidators.filter {
```
ERussel, line 337: This is will not work if we select a validator from the search and it is not in the elected list

PR #807, opened 2023-08-30
Fix parachain flow & search colator/validators

ERussel:
- fix parachain staking search from start staking flow #85ztu8tee
- remove hints from parachain start staking confirm screen #85ztuga2h
- fix validator search for case when found validator is not including to the final list
- improve pool search on bad connection #85ztvqebj


stepanLav:
Task linked: [CU-85ztvqebj iOS. Search can not be loaded on bad internet](https://app.clickup.com/t/85ztvqebj)

lynx56 (approved):

PR #811, opened 2023-09-03
Nomination Pools. Transaction history and reward details screen

lynx56:
- Added parsing poolReward type for history response
- Subtitle "Pool Rewards" for history item
- Added view for Pool Reward details

history

nova-pool-staking


novawallet/Common/Network/Subquery/SubqueryHistoryOperationFactory.swift
```diff
@@ -65,7 +65,13 @@ final class SubqueryHistoryOperationFactory {
         }
 
         if filter.contains(.rewardsAndSlashes) {
-            filterStrings.append("{ reward: { isNull: false } }")
+            let filter = SubqueryInnerFilter(inner:
+                SubqueryCompoundFilter.or([
+                    SubqueryIsNotNullFilter(fieldName: "reward"),
+                    SubqueryIsNotNullFilter(fieldName: "poolReward")
```
valentunn, line 71: This will cause request to fail on networks without nominations pools. We faced the same bug on Andrid

novawallet/Common/Network/Subquery/Models/SubqueryHistory.swift
```diff
@@ -52,6 +59,7 @@ struct SubqueryHistoryElement: Decodable {
         case extrinsic
         case transfer
         case assetTransfer
+        case poolReward
```
valentunn, line 62: Note that we should ensure this key wont be included in request for networks without a pool, otherwise request will fail

PR #819, opened 2023-09-13
Detect staking on start staking flow

ERussel:
- detect if staking have been initiated from different place during start staking flow #85ztrendq

staking-already-initiated

- take into account that MinDelegatorStk might not be available for Parachain staking - use MinDelegation in that case


stepanLav:
Task linked: [CU-85ztrendq iOS. Detect active staking on start staking flow](https://app.clickup.com/t/85ztrendq)

PR #827, opened 2023-09-18
Hide zero balances in send flow

lynx56:
- Added filter that transferable asset balance is positive
- Show transferable balance in list
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 18 33 22
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 18 33 29
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 18 33 35


novawallet/Modules/AssetsSearch/AssetOperation/Send/SendAssetsOperationInteractor.swift
```diff
@@ -0,0 +1,82 @@
...
+            logger: logger
+        )
+
+        builder?.apply(query: "")
+
+        stateObservable.addObserver(with: self) { [weak self] _, newState in
+            guard let self = self else {
+                return
+            }
+            self.builder?.apply(state: self.createState(from: newState.value))
```
ERussel, line 57: I think we must separate builder internal model from the state model. Currently it is hard to track the logic when we take a model from the state modify it and put to another builder that actually doesn't need part of it. Meaning, we passing around balances only for the purpose to model the same state model and pass to the builder.

ERussel (changes requested):
Empty state is missing for send search:

search-empty-state

novawallet/Modules/AssetsSearch/Model/AssetSearchBuilder.swift
```diff
@@ -1,36 +1,43 @@
...
     let workingQueue: DispatchQueue
     let callbackQueue: DispatchQueue
     let operationQueue: OperationQueue
     let callbackClosure: (AssetSearchBuilderResult) -> Void
     let filter: ChainAssetsFilter?
     let logger: LoggerProtocol
 
-    private var state: AssetListState
+    private var model: AssetListModel
+    private var state: AssetListState?
```
ERussel, line 14: How does this property used?

ERussel (approved):

PR #829, opened 2023-09-19
Buy tokens call to action for zero list on assets/send flow

lynx56:
- Added new cell with action for empty state
- Split factory method for send/receive/buy operations
- Made subclass Send View Controller from Search View Controller because of new empty state
- Keyboard is not appeared when screen is opening 

Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 14 24 42
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 14 24 23
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-19 at 14 24 56


novawallet/Modules/AssetsSearch/AssetOperation/AssetOperationViewFactory.swift
```diff
@@ -19,30 +17,65 @@ enum AssetOperationViewFactory {
...
-            case .buy:
-                return R.string.localizable.assetOperationBuyTitle(preferredLanguages: languages)
-            }
+            R.string.localizable.assetOperationReceiveTitle(preferredLanguages: $0.rLanguages)
         }
 
         let view = AssetsSearchViewController(
             presenter: presenter,
-            keyboardAppearanceStrategy: ModalNavigationKeyboardStrategy(),
+            keyboardAppearanceStrategy: EventDrivenKeyboardStrategy(events: Set()),
```
ERussel, line 78: Why do we need to change this strategy?

novawallet/Modules/AssetList/View/AssetListFlowLayout.swift
```diff
@@ -8,7 +8,7 @@ enum AssetListMeasurement {
     static let nftsHeight = 56.0
     static let assetHeight: CGFloat = 56.0
     static let assetHeaderHeight: CGFloat = 45.0
-    static let emptyStateCellHeight: CGFloat = 198
+    static let emptyStateCellHeight: CGFloat = 230
```
ERussel, line 11: Probably we should try to set the minimum height and then scale based on the context

novawallet/Modules/AssetsSearch/AssetSearch/AssetsSearchPresenter.swift
```diff
@@ -44,9 +44,9 @@ class AssetsSearchPresenter: AssetsSearchPresenterProtocol {
             viewModels.isEmpty,
             !result.state.balanceResults.isEmpty,
             result.state.balanceResults.count >= result.state.allChains.count {
-            view?.didReceiveGroups(state: .empty)
+            view?.didReceiveGroups(state: .init(isFiltered: false, listState: .empty))
```
ERussel, line 47: Looks like we need another interface as search currently doesn't have filtering notion

ERussel (approved):

PR #839, opened 2023-09-26
Fix images without extension in html parser

lynx56:
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-26 at 17 21 28

### DIFF
--- a/novawallet/Common/Extension/Foundation/NSAttributedString+Helpers.swift
+++ b/novawallet/Common/Extension/Foundation/NSAttributedString+Helpers.swift
@@ -24,7 +24,7 @@ extension NSAttributedString {
         mapClosure: (TAttachment) -> TNewAttachment?
     ) -> NSAttributedString where TNewAttachment: NSTextAttachment {
         let mutableAttributedString = NSMutableAttributedString(attributedString: self)
-        let range = NSRange(location: 0, length: mutableAttributedString.string.utf16.count)
+        let range = NSRange(location: 0, length: mutableAttributedString.length)
         mutableAttributedString.enumerateAttribute(
             .attachment,
             in: range,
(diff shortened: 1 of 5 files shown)


novawallet/Common/Extension/Foundation/NSAttributedString+Helpers.swift
```diff
@@ -17,3 +18,24 @@ extension NSAttributedString {
         }
     }
 }
+
+extension NSAttributedString {
+    func replacingAttachment<TAttachment, TNewAttachment>(
+        mapClosure: (TAttachment) -> TNewAttachment?
+    ) -> NSAttributedString where TNewAttachment: NSTextAttachment {
+        let mutableAttributedString = NSMutableAttributedString(attributedString: self)
+        let range = NSRange(location: 0, length: mutableAttributedString.string.utf16.count)
```
ERussel, line 27: I think it should be NSRange(location: 0, length: mutableAttributedString.length)

novawallet/Modules/Vote/Governance/View/Html/NovaTextAttachmentable.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+extension NSTextStorage: NovaTextAttachmentable {
+    public func replace(
+        attachment: NovaImageTextAttachment,
+        with resizableAttachment: ZResizableNSTextAttachment
+    ) {
+        let attributedString = self
+        let mutableAttributedString = NSMutableAttributedString(attributedString: attributedString)
+        let range = NSRange(location: 0, length: mutableAttributedString.string.utf16.count)
```
ERussel, line 15: the same comment about range here

ERussel (changes requested):
As we now control image loading I think it make sense to use Kingfisher to cache images and apply filters for them. That will allow also to fix problem with rendering. Let's keep the branch and this later

PR #843, opened 2023-09-29
Setting full amount after unstake all for relaychain

lynx56:
Simulator Screenshot - iPhone 14 - 2023-09-29 at 08 56 04
Simulator Screenshot - iPhone 14 - 2023-09-29 at 08 56 09


novawallet/Modules/Staking/Validation/StakingDataValidatorFactory.swift
```diff
@@ -219,26 +219,55 @@ extension StakingDataValidatingFactory: StakingDataValidatingFactoryProtocol {
         })
     }
 
-    func stashIsNotKilledAfterUnbonding(
-        amount: Decimal?,
-        bonded: Decimal?,
-        minimumAmount: Decimal?,
+    func minStakeNotCrossed(
```
ERussel, line 222: Is there any reason we can't have StakingBaseValidatorFactory and have this method implementation once there? Currently it is duplicated in pools and direct staking validations as I understand.

ERussel (approved):

PR #848, opened 2023-10-04
Swaps: Setup

lynx56:
Here:
- New control
- UI without details section

Simulator Screenshot - iPhone 14 - 2023-10-04 at 08 18 17


novawallet/Modules/Swaps/Setup/View/SwapSetupViewLayout.swift
```diff
@@ -0,0 +1,71 @@
...
+            $0.height.equalTo(18)
+        }
+        addArrangedSubview(receiveAmountInputView)
+        receiveAmountInputView.snp.makeConstraints {
+            $0.height.equalTo(64)
+        }
+
+        addSubview(switchButton)
+        switchButton.snp.makeConstraints {
+            $0.height.equalTo(switchButton.snp.width)
```
ERussel, line 65: As I understand currently button's size will be equal to the size of the icon. Let's make it 44 x 44 to have more area for tap.

novawallet/Modules/Swaps/Setup/View/SwapSetupViewLayout.swift
```diff
@@ -0,0 +1,71 @@
+import UIKit
+import SoraUI
+
+final class SwapSetupViewLayout: ScrollableContainerLayoutView {
+    let payAmountView: TitleHorizontalMultiValueView = .create {
```
ERussel, line 5: I think we don't take into account the fact that Max must be a button.

novawallet/Modules/Swaps/Setup/View/SwapAssetControl.swift
```diff
@@ -0,0 +1,154 @@
+import UIKit
+import SoraUI
+
+final class SwapAssetControl: BackgroundedContentControl {
```
ERussel, line 4: Let's also make this control change opacity (to 0.5) when it is highlighted. Currently there is no feedback on tap.

ERussel (approved):
Good work overall. Probably we can merge this PR and apply changes in the "Base Layout for Setup" task.

PR #854, opened 2023-10-04
Swaps: Select token to  pay

lynx56:
Simulator Screenshot - iPhone 14 - 2023-10-05 at 00 23 11


novawallet/Modules/AssetsSearch/Swaps/SwapAssetsOperationInteractor.swift
```diff
@@ -0,0 +1,123 @@
...
+    let chainAsset: ChainAsset?
+    let assetConversionOperationFactory: AssetConversionOperationFactoryProtocol
+
+    private let operationQueue: OperationQueue
+    private var builder: AssetSearchBuilder?
+    private var directionsCall: CancellableCall?
+    private var availableDirections: [ChainAssetId: Set<ChainAssetId>]?
+
+    init(
+        stateObservable: AssetListModelObservable,
```
ERussel, line 18: Like the idea, however this means that we will need stateObservable to go through SwapSetup module.

ERussel (approved):
Didn't find a way how to test the UI. Approving based on screenshot.

novawallet/Modules/AssetsSearch/Swaps/SwapAssetListViewModelFactory.swift
```diff
@@ -0,0 +1,47 @@
...
+        case .loading:
+            return (balanceState, priceState)
+        case let .cached(value):
+            return (balanceState, .cached(value: wrap(price: value)))
+        case let .loaded(value):
+            return (balanceState, .loaded(value: wrap(price: value)))
+        }
+    }
+
+    private func wrap(price: String) -> String {
```
ERussel, line 41: Let's get rid of this otherwise the list of assets will not be consistent for example with Assets Tab. Will leave comment in mockups.

PR #856, opened 2023-10-05
Swaps: Select token to receive screen

lynx56:
Simulator Screenshot - iPhone 14 - 2023-10-05 at 12 28 15

### DIFF
--- a/novawallet/Modules/AssetsSearch/Swaps/SwapAssetsOperationViewFactory.swift
+++ b/novawallet/Modules/AssetsSearch/Swaps/SwapAssetsOperationViewFactory.swift
@@ -21,6 +21,25 @@ enum SwapAssetsOperationViewFactory {
         )
     }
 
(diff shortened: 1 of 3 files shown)


novawallet/Modules/AssetsSearch/Swaps/SwapAssetsOperationViewFactory.swift
```diff
@@ -21,6 +21,25 @@ enum SwapAssetsOperationViewFactory {
         )
     }
 
+    static func createSelectReceiveTokenView(
```
ERussel, line 24: I wouldn't name it like SelectReceive/Pay token view. I think we just can have one function that accepts optional ChainAsset model.

ERussel (approved):
I think the screenshot is not valid. As I understand we want to show a case that a user has selected first asset and want to select the second one - however all assets are on the screenshot.

PR #857, opened 2023-10-06
Swaps setup details

lynx56:


novawallet/Modules/Swaps/Setup/SwapSetupInteractor.swift
```diff
@@ -1,7 +1,181 @@
...
+extension SwapSetupInteractor: SwapSetupInteractorInputProtocol {
+    func setup() {
+        feeProxy.delegate = self
+    }
+
+    func calculateQuote(for args: AssetConversion.QuoteArgs) {
+        quote(args: args)
+    }
+
+    func set(chainModel: ChainModel) {
```
ERussel, line 135: I think we also need an interface for updating assets. Because we need to subscribe to balances for those assets. Probably, we can replace this for 2 methods: setPayAsset and setReceiveAsset. In this case we will use chain from setPayAsset to create Extrinsic Service.

novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -1,58 +1,381 @@
...
+            let payChainAsset = payChainAsset,
+            let receiveChainAsset = receiveChainAsset,
+            let payInPlank = absoluteValue(for: payAmountInput)?.toSubstrateAmount(
+                precision: Int16(payChainAsset.asset.precision)),
+            let receiveInPlank = receiveAmountInput?.toSubstrateAmount(precision: Int16(receiveChainAsset.asset.precision))
+        else {
+            return
+        }
+
+        interactor.calculateFee(for: .init(
```
ERussel, line 222: Let's put todo here as it is incomplete logic

novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -1,58 +1,381 @@
...
+            priceData: payAssetPriceData,
+            locale: selectedLocale
+        )
+
+        view?.didReceiveNetworkFee(viewModel: .loaded(value: viewModel))
+    }
+
+    private func estimateFee() {
+        guard
+            let payChainAsset = payChainAsset,
```
ERussel, line 213: We can just create args for fee from quote result.

novawallet/Modules/Swaps/Setup/SwapSetupWireframe.swift
```diff
@@ -1,3 +1,37 @@
...
 
-final class SwapSetupWireframe: SwapSetupWireframeProtocol {}
+final class SwapSetupWireframe: SwapSetupWireframeProtocol {
+    let assetListObservable: AssetListModelObservable
+
+    init(assetListObservable: AssetListModelObservable) {
+        self.assetListObservable = assetListObservable
+    }
+
+    func showPayTokenSelection(
```
ERussel, line 10: We need to present token selection modally

novawallet/Common/View/StackTable/StackTitleMultiValueEditCell.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+import UIKit
+
+final class StackTitleMultiValueEditCell: RowView<GenericTitleValueView<IconDetailsView, GenericPairValueView<IconDetailsView, UILabel>>> {
```
ERussel, line 4: The problem with this implementation is that we must have some tappable UIControl in place of IconDetailsView. Probably, just user RoundedButton or UIButton

novawallet/Modules/Swaps/Setup/View/SwapSetupViewLayout.swift
```diff
@@ -27,6 +27,26 @@ final class SwapSetupViewLayout: ScrollableContainerLayoutView {
         $0.imageWithTitleView?.iconImage = R.image.iconActionSwap()
     }
 
+    let detailsHeaderCell: CollapsableViewHeader = .create {
```
ERussel, line 30: I think we can just implement the behaviour with collapsing using the same approach as in NetworkInfoView that is used in Staking. But I would use StackTableView there instead of UIStackView that is used in NetworkInfoView.

ERussel (approved):

ERussel (commented):
Let's address comments before merging

ERussel (changes requested):
1) Store as state in presenter: QuoteArgs, Quote. Then,

a) If no quote args then we don't display swap details

ERussel (approved):

PR #858, opened 2023-10-11
Swaps: Max button in setup

lynx56:
Changes:

- Max button logic implementation
- Removed SwapAssetListViewModelFactory with ~ prices formatting
- Added info sheets
- Changed max button layout

Simulator Screenshot - iPhone 14 - 2023-10-11 at 22 36 56


novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -134,21 +134,29 @@ final class SwapSetupPresenter {
...
+        switch input {
+        case .rate:
+            includedFee = fee ?? 0
+        case .absolute:
+            includedFee = 0
         }
 
-        return input.absoluteValue(from: transferrableBalanceDecimal)
+        let transferable = assetBalance?.transferable ?? 0
+        let balance = max(transferable - includedFee, 0)
```
ERussel, line 152: I think the difference may overflow and crash the app. It is better to compare BigUInts.

novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -134,21 +134,29 @@ final class SwapSetupPresenter {
...
             return nil
         }
-        guard let transferrableBalanceDecimal =
-            Decimal.fromSubstrateAmount(
-                assetBalance?.transferable ?? 0,
-                precision: payChainAsset.asset.displayInfo.assetPrecision
-            ) else {
-            return nil
+        let includedFee: BigUInt
+        switch input {
```
ERussel, line 144: We don't need this switch case. Because in case of absolute value the input will use it's own internal value.

ERussel (approved):

PR #861, opened 2023-10-23
Swaps setup validations

lynx56:


ERussel (approved):

ERussel (changes requested):
Please, fix conflicts and build

ERussel (approved):

PR #866, opened 2023-10-25
Fix balance update after transfer all and crashes

ERussel:
- fix issue when all tokens are spent but the amount is not updated on the asset search/send #8692zwbwq
- fix crash when last wallet is removed from the app
- fix crash on fast open/close NftList screen
- fix crash on fast open/close YourContributions screen


stepanLav:
Task linked: [CU-8692zwbwq iOS. Not synced balance on assets list after send all](https://app.clickup.com/t/8692zwbwq)

stepanLav:
Task linked: [CU-8692zwbwq iOS. Not synced balance on assets list after send all](https://app.clickup.com/t/8692zwbwq)

lynx56 (approved):

PR #878, opened 2023-11-08
Add validations and improvements

ERussel:
- add field validations on setup amount: zero balance, not enough balance, no liquidity, min balance violation #8692tuvcf
- add validations on continue #8692tuvcf #86932034k
- refresh quote each block
- introduce single model for slippage and config for it #86931713q
- introduce base presenter to share common logic between setup and confirm
- improve state transition #86931t73z


stepanLav:
Task linked: [CU-8692tuvcf iOS. Validations on Continue](https://app.clickup.com/t/8692tuvcf)

lynx56 (approved):

PR #882, opened 2023-11-10
Swaps: Repeat operation from history

lynx56:


novawallet/Modules/AssetDetails/AssetDetailsViewFactory.swift
```diff
@@ -5,7 +5,8 @@ struct AssetDetailsViewFactory {
     static func createView(
         assetListObservable: AssetListModelObservable,
         chain: ChainModel,
-        asset: AssetModel
+        asset: AssetModel,
+        swapCompletionClosure: SwapCompletionClosure?
```
ERussel, line 9: I would wrap assetListObservable and swapCompletionClosure into struct AssetOperationState

novawallet/Modules/OperationDetails/OperationDetailsPresenter.swift
```diff
@@ -165,7 +165,24 @@ extension OperationDetailsPresenter: OperationDetailsPresenterProtocol {
     }
 
     func repeatOperation() {
```
ERussel, line 167: I think we can reuse send action. The idea is that we have only one handler for the main action that switch cases between models and calls appropriate handler

novawallet/Modules/Swaps/Confirm/SwapConfirmPresenter.swift
```diff
@@ -425,8 +425,15 @@ extension SwapConfirmPresenter: SwapConfirmInteractorOutputProtocol {
     }
 
     func didReceiveConfirmation(hash _: String) {
+        guard let payChainAsset = getPayChainAsset() else {
```
ERussel, line 428: We first need to let view stop loading

ERussel (approved):

PR #884, opened 2023-11-13
Bugfix: Constant fee loading for low liquidity validation

lynx56:
combine_images-15

### DIFF
--- a/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
+++ b/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
@@ -11,7 +11,7 @@ final class SwapSetupPresenter: SwapBasePresenter {
 
     private(set) var quoteArgs: AssetConversion.QuoteArgs? {
         didSet {
-            provideDetailsViewModel(isAvailable: quoteArgs != nil)
+            provideDetailsViewModel()
         }
     }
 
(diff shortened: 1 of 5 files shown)


novawallet/Modules/Swaps/Setup/SwapSetupViewController.swift
```diff
@@ -200,7 +209,8 @@ extension SwapSetupViewController: SwapSetupViewProtocol {
         switch viewModel {
         case let .asset(assetViewModel):
             rootView.payAmountInputView.bind(assetViewModel: assetViewModel)
-            rootView.depositTokenButton.imageWithTitleView?.title = "Get \(assetViewModel.symbol)"
+            depositTokenSymbol = assetViewModel.symbol
```
ERussel, line 212: Thank for fixing this. I think we just need to apply current localisation without saving symbol locally. In case localisation change presenter will provide new view model.

novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -25,6 +25,15 @@ final class SwapSetupPresenter: SwapBasePresenter {
...
     private var feeIdentifier: SwapSetupFeeIdentifier?
     private var slippage: BigRational
+    private var issues: [SwapSetupViewIssue] = [] {
+        didSet {
+            provideDetailsViewModel()
+        }
+    }
+
+    private var detailsAvailable: Bool {
+        !issues.contains(.noLiqudity) && quoteArgs != nil
```
ERussel, line 35: Better to check that quoteResult is not error. In this case we also don't need save issues locally

stepanLav:
Task linked: [CU-86933nyy3 iOS. Constant fee loading for low liquidity validation](https://app.clickup.com/t/86933nyy3)

stepanLav:
Task linked: [CU-86933qjbc iOS. Can swap into zero amount](https://app.clickup.com/t/86933qjbc)

PR #885, opened 2023-11-14
Check balances/network availability for swap feature

lynx56:


novawallet/Modules/AssetList/Models/AssetListBuilderResult.swift
```diff
@@ -50,7 +50,16 @@ struct AssetListBuilderResult {
         }
 
         func hasSwaps() -> Bool {
-            allChains.values.contains { $0.hasSwaps }
+            balanceResults.contains {
```
ERussel, line 53: As we have much less chains then the balances, then I would iterate over the chains

ERussel (approved):

PR #886, opened 2023-11-14
BugFix: Auto switch to self-sufficient token if no way to pay fee in native token

lynx56:


novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -601,6 +604,22 @@ extension SwapSetupPresenter {
...
         provideNotification()
     }
+
+    private func switchFeeChainAssetIfNecessary() {
+        guard
+            !isManualFeeSet,
+            let payChainAsset = getPayChainAsset(),
+            let feeChainAsset = getFeeChainAsset(),
+            feeChainAsset.chainAssetId == payChainAsset.chain.utilityChainAssetId(),
+            let feeAssetBalance = feeAssetBalance,
```
ERussel, line 614: let's also check that we have transferrable > 0 for payChainAsset

ERussel (approved):

PR #887, opened 2023-11-16
Batch of bugfixes

lynx56:
Bugfixes:

- Apply sorting rules to token to pay/receive screens. (Sorting by fiat, then by planks, then by default)
- Add frames to validations (Always show red frame if text field contains error)
- Hide price difference for zero amounts
- Update info icon through whole app
- Add Wiki to Settings
- WalletConnect. Can't scroll network list

I deleted RewardEstimationView.swift because it's not used in app
Also fixed cell selection in swap details


novawallet/ru.lproj/Localizable.strings
```diff
@@ -114,6 +114,7 @@
 "settings.preferences" = "Предпочтения";
 "settings.security" = "Безопасность";
 "settings.community" = "Сообщество";
+"settings.wiki" = "Википедия";
```
ERussel, line 117: Вики

novawallet/Common/View/CollapsableContainerView.swift
```diff
@@ -115,9 +114,7 @@ class CollapsableContainerView: UIView {
 
         contentView.addSubview(stackView)
         stackView.snp.makeConstraints { make in
-            make.leading.trailing.equalToSuperview()
-            make.top.equalToSuperview()
-            make.bottom.equalToSuperview().inset(Constants.stackViewBottomInset)
+            make.edges.equalToSuperview()
```
ERussel, line 117: Why do we need to change this layout?

novawallet/Modules/AssetList/Models/AssetListModelHelpers.swift
```diff
@@ -19,23 +19,24 @@ enum AssetListModelHelpers {
         from chain: ChainModel,
         assets: [AssetListAssetModel]
     ) -> AssetListGroupModel {
-        let value: Decimal = assets.reduce(0) { result, asset in
-            result + (asset.totalValue ?? 0)
+        let amountValue: AmountPair<BigUInt, Decimal> = assets.reduce(.init(amount: 0, value: 0)) { result, asset in
```
ERussel, line 22: This this changes sorting on all screens that contains assets list. As I understand in terms of the task we need only check sorting on the Swap Pay/Receive screens.

lynx56: same sort used in android app for common list

novawallet/Common/View/CollapsableContainerView.swift
lynx56: because we don't need bottom inset

ERussel (approved):

PR #890, opened 2023-11-19
Realtime history updates 86935bh7e

ERussel:
- save swap locally after it sent to blockchain
- parse block events and match swap extrinsic to update local state


novawallet/Common/Services/PersistExtrinsicService/PersistTransferDetails.swift
```diff
@@ -16,3 +16,16 @@ struct PersistExtrinsicDetails {
     let callPath: CallCodingPath
     let fee: BigUInt?
 }
+
+struct PersistSwapDetails {
+    let txHash: Data
+    let sender: AccountAddress
+    let receive: AccountAddress
```
lynx56, line 23: receiver?

lynx56 (approved):

PR #895, opened 2023-11-22
Deeplinks: Open dapp screen by deeplink #86934jpye 

lynx56:
Action: xcrun simctl openurl booted "novawallet://nova/open//dapp?url="https://polkadot.polkassembly.io/""
combine_images-18

### DIFF
--- a/novawallet/Common/URLHandling/Parsing/OpenDAppUrlParsingService.swift
+++ b/novawallet/Common/URLHandling/Parsing/OpenDAppUrlParsingService.swift
@@ -1,0 +1,66 @@
+import Foundation
+import RobinHood
+
+final class OpenDAppUrlParsingService: OpenScreenUrlParsingServiceProtocol, AnyCancellableCleaning {
+    private let dAppProvider: AnySingleValueProvider<DAppList>
+    private var fetchListCall: CancellableCall?
+
+    enum Key {
+        static let url = "url"
+    }
+
+    convenience init() {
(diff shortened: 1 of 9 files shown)


stepanLav:
Task linked: [CU-86934jpye iOS. Implement open screen deeplinks](https://app.clickup.com/t/86934jpye)

novawallet/Common/URLHandling/Parsing/OpenDAppUrlParsingService.swift
```diff
@@ -0,0 +1,61 @@
...
+                .caseInsensitiveCompare(Key.url) == .orderedSame
+        })?.value.map { URL(string: $0) }
+
+        guard let dAppUrl = dAppUrl else {
+            return .failure(.openDAppScreen(.invalidURL))
+        }
+
+        let semaphore = DispatchSemaphore(value: 0)
+
+        var result: Result<UrlHandlingScreen, DeeplinkParseError>?
```
ERussel, line 41: I think it is better to pass url to the DAppListPresenter and then wait when metadata comes from the interactor. Similarly what you did with governance. The main reason is that we are blocking main thread by waiting the database. Also we might run into the situation when the database is not up to date.

ERussel: Let's fix the issue in gov PR

PR #935, opened 2024-01-11
Setup staking proxy

lynx56:
combine_images-28


novawallet/Common/Services/RemoteSubscription/Substrate/StakingAccountSubscription.swift
```diff
@@ -130,6 +130,8 @@ final class StakingAccountSubscription: WebSocketSubscribing {
             )
         }
 
+        requests.append(.init(storagePath: Proxy.proxyList, accountId: stashId))
```
ERussel, line 133: Check is chain supports proxy

novawallet/Modules/Staking/StakingMain/Relaychain/StakingRelaychainPresenter.swift
```diff
@@ -342,6 +342,11 @@ extension StakingRelaychainPresenter: StakingMainChildPresenterProtocol {
                 let stashAddress = validatorState.stashItem.stash
                 wireframe.showYourValidatorInfo(stashAddress, from: view)
             }
+        case .addProxy:
+            wireframe.showAddProxies(from: view)
+        case .editProxies:
+            // TODO: show proxy list
+            fallthrough
```
ERussel, line 349: I would add wireframe method and put todo there

novawallet/Modules/Swaps/Base/View/SwapNetworkFeeViewCell.swift
```diff
@@ -1,11 +1,11 @@
 import SoraUI
 
-final class SwapNetworkFeeViewCell: RowView<SwapNetworkFeeView>, StackTableViewCellProtocol {
+final class SwapNetworkFeeViewCell: RowView<NetworkFeeInfoView>, StackTableViewCellProtocol {
```
ERussel, line 3: What is the purpose to change the viewModel?

lynx56: I just renamed class SwapNetworkFeeView  because now it is no longer exclusively used in Swaps; it is also using in Staking.

novawallet/Common/DataProvider/Subscription/Proxy/ProxyListLocalSubscriptionFactory.swift
```diff
@@ -2,24 +2,40 @@ import RobinHood
...
 protocol ProxyListLocalSubscriptionFactoryProtocol {
     func getProxyListProvider() throws -> StreamableProvider<ProxyAccountModel>
+    func getProxyListProvider(
+        for accountId: AccountId,
+        chainId: ChainModel.Id
+    ) throws -> AnyDataProvider<DecodedProxyDefinition>
 }
 
 final class ProxyListLocalSubscriptionFactory: BaseLocalSubscriptionFactory {
     static let shared = ProxyListLocalSubscriptionFactory(
```
ERussel, line 12: Let's not make it shared because it used only in staking and no sense to store otherwise. I would store it in the state as StakingLocalSubscriptionFactory

ERussel (approved):

novawallet/Common/DataProvider/Subscription/Proxy/ProxyListLocalSubscriptionFactory.swift
lynx56: It's also used in walletNotificationService

PR #942, opened 2024-01-19
Fix proxy signing bottom sheet and logic

ERussel:
- fix identity parsing from new runtime
- fix colouring on signing bottom sheet 8693hwqy3
- save skip confirmation per proxied 8693hykyf


novawallet/Modules/Proxy/ProxyMessageSheet/ProxyMessageSheetViewFactory.swift
```diff
@@ -62,21 +80,28 @@ enum ProxyMessageSheetViewFactory {
...
         let title = LocalizableResource { locale in
             R.string.localizable.proxySigningNotEnoughPermissionsTitle(preferredLanguages: locale.rLanguages)
         }
 
         let message = LocalizableResource { locale in
-            R.string.localizable.proxySigningNotEnoughPermissionsMessage(
-                proxiedName,
-                proxyName,
-                type.value(for: locale),
+            let marker = AttributedReplacementStringDecorator.marker
```
lynx56, line 88: why?

ERussel: This is how the decorator works: we first replace formats with the marker and then dermine positions of the marker to highlight replacements

PR #950, opened 2024-01-23
Staking revoke proxy access

lynx56:
combine_images-32


novawallet/Modules/Staking/StakingProxy/Management/StakingProxyManagementPresenter.swift
```diff
@@ -74,17 +74,24 @@ final class StakingProxyManagementPresenter {
 
             return StakingProxyManagementViewModel(
                 info: info,
-                accountId: definition.proxy
+                account: .init(
```
ERussel, line 77: Why do we need the whole model in the viewModel?

ERussel (approved):

PR #959, opened 2024-01-29
Synchronization of proxies for a network after adding/removing proxies in staking

lynx56:


novawallet/Common/Services/RemoteSubscription/Substrate/ProxyAccountSubscription.swift
```diff
@@ -0,0 +1,200 @@
...
+
+            let updateClosure: (StorageSubscriptionUpdate) -> Void = { [weak self] update in
+                self?.handleUpdate(update.params.result)
+            }
+
+            let failureClosure: (Error, Bool) -> Void = { [weak self] error, unsubscribed in
+                self?.logger?.error("Did receive subscription error: \(error) \(unsubscribed)")
+            }
+
+            let subscriptionId = try connection.subscribe(
```
ERussel, line 157: It is a complex legacy way of doing subscription. It should work for now but i came up with simpler one using ```CallbackBatchStorageSubscription```. You can check it in the ```RelaychainMultistakingUpdateService```.

ERussel (approved):

PR #964, opened 2024-02-05
Hydra UI integration & refactoring

ERussel:
During integrating Hydra into the Swaps UI found the things that improved:

- there are quote and extrinsic parameters that we need to get onchain and share between Setup and Confirm screens. For Hydra those are implemented in ```HydraOmnipoolQuoteParamsService``` and ```HydraOmnipoolSwapParamsService```. So, we need a shared state to pass between Setup and Confirm screen - for Hydra it is ```HydraOmnipoolFlowState``` and for AssetHub - ```AssetHubFlowState```. To abstract the implementation ```AssetConversionFlowStateFacade``` class is introduced.

- another thing is that different swaps logic needs to rely on different logic to trigger fee/quote recalculation. So, the legacy requote logic by block is moved to ```AssetHubRequoteService```. And requote subscription is implemented in the SwapsSetupInteractor as ```setupRequoteSubscription``` function.

Beside the refactoring this PR includes adding support for hydra-swaps option in the ```ChainModel```


novawallet/Modules/AssetConversion/Service/HydraDx/HydraDxTokenConverter.swift
```diff
@@ -2,12 +2,12 @@ import Foundation
...
 
 extension HydraDx {
-    struct LocalRemoteAssetId: Equatable {
+    struct LocalRemoteAssetId: Equatable, Hashable {
         let localAssetId: ChainAssetId
         let remoteAssetId: HydraDx.OmniPoolAssetId
     }
 
-    struct SwapPair {
+    struct SwapPair: Equatable, Hashable {
```
lynx56, line 10: Hashable includes Equatable

lynx56 (approved):

novawallet/Modules/AssetConversion/Service/HydraDx/HydraDxTokenConverter.swift
ERussel: Thanks. Will double check this in the next PR

PR #967, opened 2024-02-06
Push notification service  setup

lynx56:


novawallet/Common/Services/PushNotificationsSettingsService/PushSettings.swift
```diff
@@ -0,0 +1,82 @@
...
+
+struct PushSettings: Codable, Equatable {
+    struct Wallet: Codable, Equatable {
+        let baseSubstrate: AccountAddress
+        let baseEthereum: AccountAddress?
+        let chainSpecific: [String: String]
+    }
+
+    struct Notifications: Codable, Equatable {
+        let announcements: Bool
```
ERussel, line 15: I don't think we need to store announcements flag on the Firestore. It is stored only locally. Probably better to have a separate remote model.

ERussel (commented):
I would do the following:

1) Store announcements enabling flag in the UserDefaults since this is pure local flag that guides the app whether to subscribe to the Push Notifications via topics.

2) Implement Web3AlertsSyncService in a way that:

    - It is inherited from BaseSyncService

     - has sync logic that triggers after setup/syncUp and fetches local data -> if updatedAt shows that we need to update document send local document to firestore (I would just sent the whole document to update);

     - has update logic that saves a new document locally and sends it to the firestore.

novawallet/AppDelegate.swift
```diff
@@ -1,4 +1,5 @@
 import UIKit
+import SoraKeystore
```
ERussel, line 2: Looks like we need need it anymore

novawallet/Common/Services/Web3AlertService/RemotePushSettings.swift
```diff
@@ -0,0 +1,78 @@
+import FirebaseCore
+import FirebaseFirestore
+import SubstrateSdk
+import RobinHood
+
+typealias ChainSelection = RemotePushSettings.Selection<[String]>
```
ERussel, line 6: I would either name it RemotePushChainSelection or put inside RemotePushSettings. Because now it is quite hard to guess meaning of the type.

novawallet/Common/Services/Web3AlertService/RemotePushSettings.swift
```diff
@@ -0,0 +1,78 @@
...
+
+struct Web3AlertWallet: Codable, Equatable {
+    let baseSubstrate: AccountAddress?
+    let baseEthereum: AccountAddress?
+    let chainSpecific: [String: String]
```
ERussel, line 11: I would make Web3AllertWallet.ChainId type and also used AccountAddress for the value.

novawalletTests/Modules/Settings/SettingsTests.swift
```diff
@@ -77,6 +77,15 @@ final class SettingsTests: XCTestCase {
                 completion(nil)
             }
         }
+        
+        let operationQueue = OperationManagerFacade.sharedDefaultQueue
+        let web3AlertsSyncService = Web3AlertsSyncServiceFactory(
```
ERussel, line 82: It is assumed that unit tests doesn't make any remote requests or communicate with local database because this will lead to not deterministic behaviour.  So, let's pass stubs for web3alert and push notification services.

novawallet/Modules/Settings/SettingsInteractor.swift
```diff
@@ -16,6 +16,9 @@ final class SettingsInteractor {
     let biometryAuth: BiometryAuthProtocol
     let walletConnect: WalletConnectDelegateInputProtocol
     let walletNotificationService: WalletNotificationServiceProtocol
+    let alertNotificationsService: Web3AlertsSyncServiceProtocol
+    let operationQueue: OperationQueue
+    let pushNotificationsService: PushNotificationsServiceProtocol
```
ERussel, line 21: As I understand push notification contains web3alert sync service. Why do we need both of them here?

Also we might pass factories or proxies instead of real services because we want to start services only if push notification enabled

novawallet/Common/Services/Web3AlertService/Web3AlertsSyncService.swift
```diff
@@ -0,0 +1,174 @@
...
+        settingsManager.pushSettingsDocumentId = settings.identifier
+        return localSaveOperation(settings: settings)
+    }
+
+    func update(token: String) -> CompoundOperationWrapper<Void?> {
+        guard let documentUUID = settingsManager.pushSettingsDocumentId else {
+            return .createWithError(Web3AlertsSyncServiceError.documentNotFound)
+        }
+
+        FirebaseHolder.shared.configureApp()
```
ERussel, line 156: We have several places in file where we do this. Probably better to move to the service factory before service creation and assume it called before service created.

novawallet/Common/Services/Web3AlertService/Web3AlertsSyncService.swift
```diff
@@ -0,0 +1,174 @@
...
+import FirebaseFirestore
+import FirebaseMessaging
+
+enum Web3AlertsSyncServiceError: Error {
+    case documentNotFound
+}
+
+protocol Web3AlertsSyncServiceProtocol: ApplicationServiceProtocol {
+    func save(settings: LocalPushSettings) -> BaseOperation<Void>
+    func update(token: String) -> CompoundOperationWrapper<Void?>
```
ERussel, line 14: Interface probably must accept completion closure instead of returning wrappers. Currently the problem is that nobody tracking currently executing operation. As we discussed we don't want to execute several operations at the same time to the remote. Thus we need to save locally current executing wrapper, clear it when it is completed don't sync if we already executing something and addDependency to it if we updating/saving.

novawalletTests/Common/Services/Web3NameIntegrityVerifierTests.swift
```diff
@@ -1,4 +1,5 @@
 import XCTest
+import Foundation
```
ERussel, line 2: Don't see the reason why we adding this

ERussel (approved):

ERussel (approved):

PR #973, opened 2024-02-12
Fix/firebase_token

stepanLav:
Changes:
https://github.com/wzieba/Firebase-Distribution-Github-Action/wiki/FIREBASE_TOKEN-migration


leohar (approved):

ERussel (commented):
Will generate the service account and add pk to the secrets first

ERussel (approved):

PR #987, opened 2024-02-21
Push notifications settings: governance, staking

lynx56:


novawallet/Modules/Notifications/GovernanceNotifications/GovernanceNotificationsPresenter.swift
```diff
@@ -0,0 +1,126 @@
...
+                    identifier: $0.identifier,
+                    enabled: false,
+                    icon: RemoteImageViewModel(url: $0.icon),
+                    name: $0.name,
+                    newReferendum: true,
+                    referendumUpdate: true,
+                    delegateHasVoted: true,
+                    tracks: .all
+                )
+                settings[$0.identifier] = newModel
```
ERussel, line 50: I think it is better populate settings in the separate function and don't break functional paradigm with side effects

novawallet/Modules/Notifications/GovernanceNotifications/GovernanceNotificationsPresenter.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    private func provideClearButtonState() {
+        let isEnabled = settings.contains(where: { $0.value.enabled })
+        view?.didReceive(isClearActionAvailabe: isEnabled)
+    }
+
+    private func disableChainNotificationsIfNeeded(network: ChainModel.Id) {
+        if settings[network]?.allNotificationsIsOff == true {
+            changeSettings(chainId: network, isEnabled: false)
+            settings[network].map {
```
ERussel, line 66: Let's use if here and don't break functional paradigm with side effects

novawallet/Modules/Notifications/GovernanceNotifications/GovernanceNotificationsViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+                isOn: model.delegateHasVoted,
+                action: { self.presenter.changeSettings(
+                    chainId: model.identifier,
+                    delegateHasVoted: $0
+                ) }
+            )),
+            .accessoryCell(.init(
+                title: tracks,
+                accessory: tracksSubtitle(from: model.tracks),
+                action: { self.presenter.selectTracks(chainId: model.identifier) }
```
ERussel, line 83: I think we are having circular dependency here because self is inherited from ChainsNotificationsViewControllers which has reference to models.

novawallet/Modules/Notifications/GovernanceTracksSettings/GovernanceTracksSettingsInteractor.swift
```diff
@@ -0,0 +1,48 @@
...
+        operationQueue.addOperations(wrapper.allOperations, waitUntilFinished: false)
+    }
+}
+
+extension GovernanceTracksSettingsInteractor: GovernanceSelectTracksInteractorInputProtocol {
+    func setup() {
+        provideTracks()
+    }
+
+    func remakeSubscriptions() {}
```
ERussel, line 43: Is not used? Let's delete

novawallet/Modules/Notifications/GovernanceTracksSettings/GovernanceTracksSettingsWireframe.swift
```diff
@@ -0,0 +1,19 @@
...
+
+    func proceed(
+        from _: ControllerBackedProtocol?,
+        tracks: [GovernanceTrackInfoLocal],
+        totalCount: Int
+    ) {
+        completion(Set(tracks.map(\.trackId)), totalCount)
+    }
+
+    func proceed(from _: ControllerBackedProtocol?, tracks _: [GovernanceTrackInfoLocal]) {}
```
ERussel, line 18: Looks like not used. Let's remove

novawallet/Modules/Notifications/Management/NotificationsManagementInteractor.swift
```diff
@@ -67,24 +75,40 @@ extension NotificationsManagementInteractor: NotificationsManagementInteractorIn
...
-                self?.presenter?.didReceiveSaveCompletion()
-            }
+            group.leave()
+        }
+
+        if topicService == nil {
+            topicService = alertServiceFactory.createTopicService()
+        }
+
+        topicService?.save(settings: topics) {
```
ERussel, line 100: if topic service is nil then we don't leave. I would probably move group.enter to the if statement.

lynx56: we create topicService before

PR #998, opened 2024-03-07
Refactor notification service

ERussel:
- introduce ```PushNotificationsServiceFacade``` that allows to subscribe to the notifications state and save settings;
- separate ```Web3Alert.RemoteSettings``` and ```Web3Alert.LocalSettings``` because we store additionaly metaId for locally and need to convert between local chain id and remote chain id;
- improve ```PushNotificationsTopicService``` it now allow to ```save``` topics (by saving them locally and subscribe remotely) and ```saveDiff``` (first calculating diff between new settings and old one to find what to subscribe/unsubscribe remotely);
- improve ```PushNotificationsStatusService``` it now manages registration/deregistration for the push notifications and the state of the process; It is assumed that apns token is accepted outside (from AppDelegate in our case); Also it allows to wait until we not only register push notifications but also receive the tokens (apns and fcm) - because we can't subscribe topics before we get the tokens;


novawallet/Common/Operation/OperatingCall.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import RobinHood
+
+protocol OperatingCall: CancellableCall {
+    var allOperations: [Operation] { get }
+}
+
+extension BaseOperation: OperatingCall {
+    var allOperations: [Operation] {
+        [self]
```
lynx56, line 10: May be we should add dependencies too

lynx56 (approved):

PR #1001, opened 2024-03-11
Open screen by push

lynx56:


novawallet/Common/PushHandling/OpenPushScreenService.swift
```diff
@@ -0,0 +1,96 @@
...
+    func didAskScreenOpen(_ screen: PushHandlingScreen)
+}
+
+enum PushHandlingScreen {
+    case gov(Referenda.ReferendumIndex)
+    case historyDetails(ChainAssetId)
+    case error(Error)
+}
+
+protocol NotificationMessageHandlerProtocol: AnyObject {
```
ERussel, line 13: PushNotificationMessageHandling

novawallet/Common/PushHandling/OpenPushScreenService.swift
```diff
@@ -0,0 +1,96 @@
...
+    func handle(
+        message: NotificationMessage,
+        completion: @escaping (Result<PushHandlingScreen, Error>) -> Void
+    )
+    func cancel()
+}
+
+protocol OpenPushScreenServiceProtocol: PushHandlingServiceProtocol {
```
ERussel, line 21: PushNotificationOpenScreenProtocol

novawallet/Common/PushHandling/OpenPushScreenService.swift
```diff
@@ -0,0 +1,96 @@
...
+    var delegate: PushScreenOpenDelegate? { get set }
+
+    func consumePendingScreenOpen() -> PushHandlingScreen?
+}
+
+final class OpenPushScreenService {
```
ERussel, line 27: PushNotificationOpenScreenFacade

novawallet/Modules/AssetList/AssetListPresenter.swift
```diff
@@ -22,6 +22,7 @@ final class AssetListPresenter {
     private var hidesZeroBalances: Bool?
     private var shouldShowPolkadotPromotion: Bool = true
     private var hasWalletsUpdates: Bool = false
+    private var selectedChainAssetId: ChainAssetId?
```
ERussel, line 25: I don't like this approach. It makes AssetListPresenter much complicated. I would stick with the approach when MainTabBar just pops current details and pushes new one.

Also we don't have here the similar issue we had with governance when referendumId is not available until we fetch it from the chain. Here we already know the chain as we subscribe to it in the AssetDetailsNotificationHandler.

novawallet/Modules/MainTabBar/MainTabBarWireframe.swift
```diff
@@ -63,6 +58,63 @@ final class MainTabBarWireframe: MainTabBarWireframeProtocol {
...
+            chain: chainAsset.chain,
+            asset: chainAsset.asset,
+            operationState: operationState
+        ) else {
+            return
+        }
+
+        let navigationController = NovaNavigationController(rootViewController: detailsView.controller)
+
+        controller.present(navigationController, animated: true)
```
ERussel, line 115: Why modally? As I see we are pushing AssetDetails to the navigation stack

ERussel (approved):

PR #1023, opened 2024-03-27
Push notifications

ERussel:
- implement push notifications support via Firebase SDK
- implement a separate extension to enrich the push notification before display
- implement UI for choosing wallets and events to track


valentunn (approved):

antonkhvorov (approved):
lets gooooo

PR #1061, opened 2024-05-13
Feature/manual backup flow

svojsu:
- implement UI and view model logic for Manual Backup Wallet List screen

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-13 at 14 18 26


novawallet/Modules/ManualBackupWalletList/ManualBackupWalletListProtocols.swift
```diff
@@ -0,0 +1,11 @@
+protocol ManualBackupWalletListPresenterProtocol: WalletsListPresenterProtocol {
+    func title() -> String
```
ERussel, line 2: I think it is too complicated for concrete implementation of the list. I would just call R from View Controller to get the title.

ERussel (approved):

PR #1077, opened 2024-05-22
Feature/manual backup key list

svojsu:
Backup key selection screen added

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-22 at 17 55 16


novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+        interactor.setup()
+
+        guard let walletViewModel = try? walletViewModelFactory.createDisplayViewModel(from: metaAccount) else {
+            return
+        }
+
+        view?.updateNavbar(with: walletViewModel)
+    }
+
+    func didTapDefaultKey() {
```
ERussel, line 48: We usually use ```did``` prefix for output protocols. For presenter we need to perform action, so let's rename to ``activateDefaultKey``` and ```activateCustomKey```

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+                accounts: customChainsViewModels
+            )
+        )
+    }
+
+    func sorted(
+        _ chains: [ChainModel.Id: ChainModel],
+        for metaAccount: MetaAccountModel
+    ) -> SortedChains {
+        let defaultChainsIds = Set(metaAccount.chainAccounts.map(\.chainId))
```
ERussel, line 144: Why it is called defaultChainsIds if it holds custom chain ids?

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+            defaultChains: defaultChains,
+            customChains: customChains
+        )
+    }
+
+    func formattedString(for defaultChains: [ChainModel]) -> String {
+        let chainsToMention = defaultChains.count > 1
+            ? defaultChains.prefix(2)
+            : defaultChains.prefix(1)
+        let separator = ", "
```
ERussel, line 168: String.CompoundSeparator.commaSpace

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+        )
+        let viewModel = createViewModel(from: sortedChains)
+
+        view?.update(with: viewModel)
+    }
+}
+
+// MARK: Private
+
+private extension ManualBackupKeyListPresenter {
```
ERussel, line 80: We definitely need view model factory for the view models

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+    func formattedString(for defaultChains: [ChainModel]) -> String {
+        let chainsToMention = defaultChains.count > 1
+            ? defaultChains.prefix(2)
+            : defaultChains.prefix(1)
+        let separator = ", "
+        let restCount = defaultChains.count - chainsToMention.count
+
+        return chainsToMention
+            .map(\.name)
+            .reduce(into: "") { partialResult, name in
```
ERussel, line 173: I think it is too complicated. I think we can first joins chainsToMention name by separator and then check whether we need to append + suffix.

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+        var defaultChains: [ChainModel] = []
+        var customChains: [ChainModel] = []
+
+        chains.forEach { chain in
+            defaultChainsIds.contains(chain.key)
+                ? customChains.append(chain.value)
+                : defaultChains.append(chain.value)
+        }
+
+        defaultChains.sort { ChainModelCompator.defaultComparator(chain1: $0, chain2: $1) }
```
ERussel, line 155: Let's also filter default chains and leave only those for which ```metaAccount.fetch(request: chain.accountRequest())``` is not nil.

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
```diff
@@ -0,0 +1,200 @@
...
+private extension ManualBackupKeyListPresenter {
+    func createViewModel(from sortedChains: SortedChains) -> ManualBackupKeyListViewLayout.Model {
+        let listHeaderText = R.string.localizable.chainAccountsListHeader(
+            preferredLanguages: localizationManager.selectedLocale.rLanguages
+        )
+
+        return .init(
+            listHeaderText: listHeaderText,
+            accountsSections: [
+                createDefaultChainsSection(for: sortedChains.defaultChains),
```
ERussel, line 89: We don't need to display section for the empty chains list

novawallet/Common/View/ChainAccount/ChainAccountView.swift
```diff
@@ -13,12 +13,20 @@ final class ChainAccountView: UIView {
...
-        return label
-    }()
+    let multiLabelView: MultiValueView = .create { view in
+        view.valueTop.apply(style: .regularSubhedlinePrimary)
+        view.valueTop.textAlignment = .left
+
+        view.valueBottom.apply(style: .caption1Secondary)
+        view.valueBottom.textAlignment = .left
+        view.valueBottom.numberOfLines = 0
+        view.valueBottom.isHidden
```
ERussel, line 23: Forgot to assign a value. The main drawback of that approach is that we should use bind method to set the text's because otherwise we may end up with unneeded offset due to the spacing between title and subtitle. So, let's check that we are using bind method in all places where ChainAccountView is used.

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListPresenter.swift
svojsu: My bad, forgot to rename it after few rewrites 😅

novawallet/Common/View/ChainAccount/ChainAccountView.swift
svojsu: Fixed Part with Value. To address the drawbacks associated with adding MultiValueView, we have hidden valueBottom to avoid dealing with unnecessary offsets.

novawallet/Modules/ManualBackupKeyList/ManualBackupKeyListViewModelFactory.swift
```diff
@@ -0,0 +1,107 @@
...
+                        subtitle: formattedString(for: chains)
+                    )
+                ]
+            )
+        )
+    }
+
+    func createCustomChainsSection(for chains: [ChainModel]) -> ManualBackupKeyListViewLayout.Sections {
+        let customChainsViewModels = chains
+            .compactMap { [weak self] chain -> ManualBackupKeyListViewLayout.CustomAccount? in
```
ERussel, line 68: no need to weak self when closure is non escaping

ERussel (approved):

PR #1087, opened 2024-06-03
Feature/settings network option

svojsu:
- Added `source` and `enabled` fields to the `CDChain` database model;
- Added lightweight migration to v28 for the `SubstrateDataModel`;
- Added a "Networks" option in `Settings`;
- Added the `NetworksList` module to display default and added networks.

Simulator Screenshot - iPhone 15 Pro Max - 2024-06-03 at 23 36 51


novawallet/Common/View/GenericBackgroundView.swift
```diff
@@ -0,0 +1,89 @@
+import UIKit
+import SoraUI
+
+/// Class to be used to wrap given `WrappedView` with a view that has defined background with either equal `spacing` or
+/// precisly set `contentInsets`
+/// `mode: Mode` can be used to adjust how `WrappedView` is positioned within this view
+class GenericBackgroundView<WrappedView: UIView>: RoundedView {
```
ERussel, line 7: I found that we have ```GenericBorderedView```. Probably better to use that one

novawallet/Modules/NetworksList/NetworksListInteractor.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    private func subscribeChains() {
+        chainRegistry.chainsSubscribe(
+            self,
+            runningInQueue: .main
+        ) { [weak self] changes in
+            guard let self else { return }
+            presenter?.didReceiveChains(changes: changes)
+
+            chains = changes.mergeToDict(chains)
```
ERussel, line 24: I don't think we need to accumulate chains here. I would rather subscribe to the chain state once it is new and unsubscribe when delete

novawallet/Modules/NetworksList/NetworksListViewController.swift
```diff
@@ -0,0 +1,230 @@
...
+        var snapshot = Snapshot()
+
+        viewModel.sections.forEach { section in
+            snapshot.appendSections([section])
+
+            switch section {
+            case let .networks(rows), let .banner(rows):
+                snapshot.appendItems(rows)
+
+                networkViewModels = rows.compactMap { row in
```
ERussel, line 78: I think we can just switch to manual delegate and provide cell because snapshotting only complicates things if we have a separate data source

PR #1093, opened 2024-06-11
Feature/network details

svojsu:
Network details with ability to enable/disable connection, auto-balance nodes and choose node manually
Simulator Screenshot - iPhone 15 Pro Max - 2024-06-11 at 22 25 27


novawallet/Common/Storage/EntityToModel/ChainModelMapper.swift
```diff
@@ -183,6 +190,7 @@ final class ChainModelMapper {
         }
 
         entity.nodes = Set(nodeEntities) as NSSet
+        entity.selectedNode = try nodeMapping(model.selectedNode)
```
ERussel, line 193: The problem here is that ```nodeMapping``` makes a look up in the ```nodes``` list of the current chain and if it not there it creates a new entity which is just attached to the selectedNode and not in the nodes list. So, I would make a separate logic for the selectedNode that tries to find the existing node in the nodes list and if it is not there then raise an exception

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -0,0 +1,176 @@
...
+
+// MARK: NetworkDetailsInteractorInputProtocol
+
+extension NetworkDetailsInteractor: NetworkDetailsInteractorInputProtocol {
+    func setup() {
+        connectToChainNodes()
+        subscribeChainChanges()
+    }
+
+    func toggleNetwork() {
```
ERussel, line 43: I wouldn't use toggle logic here since we executing it asynchronously. Let's instead pass a concrete value here.

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -0,0 +1,176 @@
...
+
+            return [chain.byChanging(enabled: !chain.enabled)]
+        }, {
+            []
+        })
+
+        operationQueue.addOperation(saveOperation)
+    }
+
+    func toggleConnectionMode() {
```
ERussel, line 55: I wouldn't use toggle logic here since we executing it asynchronously. Let's instead pass a concrete value here.

novawallet/Modules/NetworkDetails/NetworkDetailsViewController.swift
```diff
@@ -0,0 +1,209 @@
...
+        viewModel
+            .rows
+            .forEach { row in
+                guard case let .node(nodeModel) = row else {
+                    return
+                }
+
+                let indexPath = IndexPath(
+                    row: nodeModel.index,
+                    section: 2
```
ERussel, line 47: Let's move sections indexes to Constants for the better understanding which section is hidden behind a given index

novawallet/Modules/NetworkDetails/View/NetworkDetailsNodeView.swift
```diff
@@ -0,0 +1,169 @@
...
+        separatorView.snp.makeConstraints { make in
+            make.leading.trailing.equalToSuperview().inset(16)
+            make.bottom.top.equalToSuperview().inset(separatorView.strokeWidth)
+        }
+    }
+
+    override func setHighlighted(_ highlighted: Bool, animated: Bool) {
+        super.setHighlighted(highlighted, animated: animated)
+
+        contentDisplayView.roundedContainerView.backgroundView.fillColor = highlighted
```
ERussel, line 30: I think we can set highlightedFillColor for the RoundedView and set its isHighlighted flag

novawallet/Modules/NetworkDetails/View/NetworkDetailsNodeView.swift
```diff
@@ -0,0 +1,169 @@
...
+        roundedContainerView.contentView.sView.stackView.alignment = .leading
+        roundedContainerView.contentView.sView.spacing = Constants.stackSpacing
+
+        roundedContainerView.contentView.sView.valueBottom.spacing = Constants.stackSpacing
+        roundedContainerView.contentView.sView.valueBottom.stackView.alignment = .leading
+
+        networkStatusView.spacingBetweenLabelAndIcon = Constants.stackSpacing
+        networkStatusView.titleFont = .semiBoldCaps2
+
+        setNeedsLayout()
```
ERussel, line 144: Why do we need it here?

PR #1096, opened 2024-06-12
Feature/network details with node management

svojsu:
- Added ping measuring for node connections;
- Added node connection logic to handle chain states and connection modes;
- Implement add, edit and delete for custom node;
- Show custom nodes separately from remote syncing ones.
- The "More" option for custom nodes opens a bottom sheet with management options.
- Added logic for custom nodes that prevents them from being erased during sync.


antonkhvorov:
IBP guys are champions

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -20,31 +20,38 @@ final class NetworkDetailsInteractor {
...
     func toggleNetwork() {
         let saveOperation = repository.saveOperation({ [weak self] in
             guard let self else { return [] }
 
-            return [chain.byChanging(enabled: !chain.enabled)]
+            var updatedChain = chain.byChanging(enabled: !chain.enabled)
+
+            if chain.enabled {
+                updatedChain = updatedChain.updatingSelectedNode(with: nil)
+                updatedChain = updatedChain.updatingConnectionMode(for: .manual)
```
ERussel, line 51: Didn't get why do we need even update connection mode and selected node when chain enabled? Also is selected node is nil then manual connection doesn't make sense as we don't have a node to connect to. Am I right?

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -166,6 +197,93 @@ private extension NetworkDetailsInteractor {
...
+            let startTime = CFAbsoluteTimeGetCurrent()
+            execute(
+                operation: queryOperation,
+                inOperationQueue: queue,
+                runningCallbackIn: nil
+            ) { _ in
+                let endTime = CFAbsoluteTimeGetCurrent()
+                let ping = Int((endTime - startTime) * 1000)
+
+                resultClosure(.success(ping))
```
ERussel, line 283: I think we also need check the ping result and only provide measurement for the successful pings

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -166,6 +197,93 @@ private extension NetworkDetailsInteractor {
...
+    ) -> BaseOperation<Int> {
+        AsyncClosureOperation { resultClosure in
+            let startTime = CFAbsoluteTimeGetCurrent()
+            execute(
+                operation: queryOperation,
+                inOperationQueue: queue,
+                runningCallbackIn: nil
+            ) { _ in
+                let endTime = CFAbsoluteTimeGetCurrent()
+                let ping = Int((endTime - startTime) * 1000)
```
ERussel, line 281: I think better to use our TimeInterval.milliseconds for consistency

novawallet/Modules/NetworkDetails/NetworkDetailsPresenter.swift
```diff
@@ -114,17 +117,35 @@ private extension NetworkDetailsPresenter {
...
 extension NetworkDetailsPresenter {
-    enum ConnectionState {
+    enum ConnectionState: Equatable {
         case connecting
         case connected
-        case pinged(TimeInterval)
+        case disconnected
+        case pinged(Int)
+
+        static func == (lhs: ConnectionState, rhs: ConnectionState) -> Bool {
```
ERussel, line 136: Why do we need implement comparison manually and not use derived one?

novawallet/Modules/NetworkDetails/NetworkDetailsViewModelFactory.swift
```diff
@@ -77,7 +77,8 @@ private extension NetworkDetailsViewModelFactory {
                             ),
                             icon: nil
                         ),
-                        selectable: network.enabled
+                        selectable: network.enabled,
+                        enabled: network.chainId != KnowChainId.polkadot
```
ERussel, line 81: How do we ensure that a user can't disable Polkadot network? Can't find how do we manually disabling UISwitch for it

novawallet/Modules/NetworkDetails/NetworkDetailsViewModelFactory.swift
```diff
@@ -121,15 +123,27 @@ private extension NetworkDetailsViewModelFactory {
...
+                    R.string.localizable.networkStatusConnecting(
+                        preferredLanguages: localizationManager.selectedLocale.rLanguages
+                    ).uppercased()
+                )
+            case .disconnected:
+                .disconnected
+            case let .pinged(ping):
+                switch ping {
+                case 0 ..< 100:
+                    .pinged(.low("\(ping) MS"))
```
ERussel, line 139: Let's still use localization for it

novawallet/Modules/NetworkDetails/View/NetworkDetailsViewLayout.swift
```diff
@@ -53,16 +53,34 @@ extension NetworkDetailsViewLayout {
...
         }
 
-        enum ConnectionState {
+        enum ConnectionState: Equatable {
             case connecting(String)
             case pinged(Ping)
+            case disconnected
+            case unknown
+
+            static func == (lhs: ConnectionState, rhs: ConnectionState) -> Bool {
```
ERussel, line 62: Why do we need implement comparison manually and not use derived one?

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -166,6 +197,93 @@ private extension NetworkDetailsInteractor {
...
+
+    func disconnectNodes() {
+        nodesConnections.values.forEach { connection in
+            connection.disconnect(true)
+        }
+
+        nodesConnections = [:]
+    }
+
+    func measureNodePing(for nodeUrl: URL) {
```
ERussel, line 209: Let's at least move ping logic to the operation factory. So, the NodePingOperationFactory will have a method to create ping operation for a given chain and node

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -166,6 +197,93 @@ private extension NetworkDetailsInteractor {
...
+        ) { [weak self] result in
+
+            switch result {
+            case let .success(ping):
+                self?.presenter?.didReceive(
+                    .pinged(ping),
+                    for: nodeUrl.absoluteString
+                )
+            case let .failure(error):
+                print(error)
```
ERussel, line 248: Let's communicate the error to the presenter. Probably, in that can we can switch the node to the connecting state or discuss with the designer a new state

novawallet/Modules/NetworkDetails/NetworkDetailsViewModelFactory.swift
svojsu: there is confusing naming for viewmodel's fields. I've tried to reuse the old one, so `selectable` is actually works like `isOn` and `enabled` is actually for a whole cell ability to be interacted with. That's how we ensure user can't interact with "Enable connection" cell in case of Polkadot network.

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
svojsu: From the business logic's perspective, yes, you're right. From the design perspective, there is a matching state of connection mode for a disabled network.

It seems like we should just map the view model in the right way considering the network state and not mess with the connection mode of the actual stored chain.

image

novawallet/Common/Services/ChainRegistry/ConnectionPool/ConnectionFactory.swift
```diff
@@ -162,7 +162,11 @@ extension ConnectionFactory: ConnectionFactoryProtocol {
     }
 
     private func extractNodeUrls(from chain: ChainModel, schema: String) -> [ConnectionTLSSupport] {
-        let filteredNodes = chain.nodes.filter { $0.url.hasPrefix(schema) }
+        let filteredNodes = if case let .manual(selectedNode) = chain.connectionMode {
```
ERussel, line 165: I think we need to return empty set if schema of the selected node not matching the requested schema.

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -142,18 +165,47 @@ private extension NetworkDetailsInteractor {
...
+        } else {
+            disconnectNodes()
         }
     }
 
-    func connectToChainNodes() {
+    func connectToNodes(of chain: ChainModel) {
+        nodesConnections = [:]
+
         chain.nodes.forEach { node in
```
ERussel, line 209: I would also filter here by wss schema as we might have https nodes for Ethereum to use for one shot connection

svojsu: yep, thanks

novawallet/Modules/NetworkDetails/NetworkDetailsPresenter.swift
```diff
@@ -0,0 +1,230 @@
...
+            selectedNode = nil
+        }
+
+        connectionStates[nodeURL] = connectionState
+
+        switch connectionState {
+        case .connecting, .disconnected, .pinged, .unknown:
+            provideNodeViewModel(for: nodeURL)
+        default:
+            break
```
ERussel, line 113: What are the other options? Let's manually state them

novawallet/Modules/NetworkDetails/NetworkDetailsPresenter.swift
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    func addNode() {
+        wireframe.showAddNode(
+            from: view,
+            chainId: chain.chainId
+        )
+    }
+
+    func selectNode(at indexPath: IndexPath) {
```
ERussel, line 52: I really don't like that we are exposing index path and section based layout to the presenter. It is not making the UI flexible and harder to change it in future. We could just receive here a view model and find the corresponding model based on some id.

novawallet/Modules/NetworkManageNode/NetworkManageNodeViewController.swift
```diff
@@ -0,0 +1,111 @@
+import UIKit
+import SoraFoundation
+
+final class NetworkManageNodeViewController: UIViewController, ViewHolder {
```
ERussel, line 4: I thought we could use ```ModalPickerViewController``` for that bottomsheet. Probably we can leave it as it is for now.

novawallet/Modules/NetworkNode/Interactor/NetworkNodeAddInteractor.swift
```diff
@@ -0,0 +1,94 @@
...
+            let currentConnectingNode,
+            let chain = chainRegistry.getChain(for: chainId)
+        else { return }
+        
+        let saveOperation = repository.saveOperation(
+            { [chain.adding(node: currentConnectingNode)] },
+            { [] }
+        )
+
+        saveOperation.completionBlock = { [weak self] in
```
ERussel, line 28: Prefer ```execute(...)``` shortcut for executing operations and also handle error case

novawallet/Modules/NetworkNode/Interactor/NetworkNodeAddInteractor.swift
```diff
@@ -0,0 +1,94 @@
...
+        with url: String,
+        name: String
+    ) -> ChainNodeModel? {
+        guard let chain = chainRegistry.getChain(for: chainId) else {
+            return nil
+        }
+        
+        let currentLastIndex = chain.nodes
+            .map { $0.order }
+            .sorted { $0 < $1 }
```
ERussel, line 75: I think we can use max instead of sorting

novawallet/Modules/NetworkNode/Interactor/NetworkNodeBaseInteractor.swift
```diff
@@ -0,0 +1,151 @@
...
+                break
+            }
+        }
+    }
+}
+
+// MARK: Errors
+
+extension NetworkNodeBaseInteractor {
+    enum Errors: Error {
```
ERussel, line 123: So in this case presenter will know about NetworkNodeBaseInteractor.Errors. I would just define NetworkNodeInteractorError. And we later don't need to specify the full type when passing the error to the presenter just, for example, ```basePresenter?.didReceive(.alreadyExists(...))```

novawallet/Modules/NetworkNode/Interactor/NetworkNodeEditInteractor.swift
```diff
@@ -0,0 +1,86 @@
...
+        let saveOperation = repository.saveOperation(
+            { [weak self] in
+                guard let self else { return [] }
+                
+                return [chain.replacing(nodeToEdit, with: editedNode)]
+            },
+            { [] }
+        )
+        
+        saveOperation.completionBlock = { [weak self] in
```
ERussel, line 60: Use shortcut

novawallet/Modules/NetworkNode/NetworkNodeViewFactory.swift
```diff
@@ -0,0 +1,100 @@
...
+    static func createNodeEditView(
+        with chainId: ChainModel.Id,
+        _ node: ChainNodeModel
+    ) -> NetworkNodeViewProtocol? {
+        let connectionFactory = ConnectionFactory(
+            logger: Logger.shared,
+            operationQueue: OperationManagerFacade.assetsSyncQueue
+        )
+
+        let chainRegistry = ChainRegistryFactory.createDefaultRegistry()
```
ERussel, line 60: ```ChainRegistryFacade.sharedRegistry```. Here we are creating a new instance instead of using the singleton. Please, make sure we are not doing the same in other places.

novawallet/Modules/Staking/Services/RewardCalculatorService/Parachain/BaseParaStakingRewardCalculatoService+Fetch.swift
```diff
@@ -1,37 +1,52 @@
 import Foundation
+import SubstrateSdk
+import BigInt
+import Operation_iOS
 
 extension BaseParaStakingRewardCalculatoService {
-    func updateTotalStaked() throws {
-        totalStakedService?.throttle()
-        totalStakedService = nil
+    func updateTotalStaked() {
```
ERussel, line 7: This changes shouldn't be here. Please, make sure you have merged develop to the base branch and check that we don't have this changed.

ERussel (commented):
We need also discuss whether to allowing adding ``ws`` nodes instead of ``wss``.

novawallet/Modules/Staking/Services/RewardCalculatorService/Parachain/BaseParaStakingRewardCalculatoService+Fetch.swift
svojsu: now up to develop

novawallet/Modules/NetworkNode/Interactor/NetworkNodeBaseInteractor.swift
svojsu: agree

novawallet/Modules/NetworkNode/Interactor/NetworkNodeEditInteractor.swift
svojsu: fixed

novawallet/Modules/NetworkManageNode/NetworkManageNodeViewController.swift
svojsu: oh, haven't seen this one

novawallet/Modules/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -0,0 +1,375 @@
+import UIKit
+import SubstrateSdk
+import SoraFoundation
+import Operation_iOS
+
+final class NetworkDetailsInteractor {
+    weak var presenter: NetworkDetailsInteractorOutputProtocol?
+
+    private var chain: ChainModel
```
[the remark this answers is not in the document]
svojsu: Yes, we can pass the id instead of chain, but we need chain to operate anyway, so if we pass just the id, it will force us to check optional chain in almost every method of the class. 
We're subscribing for chain registry changes so any chain change will be processed anyway.

novawallet/Modules/NetworkNode/Interactor/NetworkNodeBaseInteractor.swift
```diff
@@ -0,0 +1,289 @@
...
+    case alreadyExists(nodeName: String)
+    case wrongNetwork(networkName: String)
+    case unableToConnect(networkName: String)
+    case wrongFormat
+    
+    case common(error: CommonError)
+}
+
+private extension NSPredicate {
+    static var ws: NSPredicate {
```
ERussel, line 246: Let's move this guy near to the NSPredicate.webSocket

ERussel (approved):

PR #1098, opened 2024-06-16
fix connection cast that always produces nil

svojsu:
Fixed the type casting of connection stored with `WeakWrapper` in the `connections` dictionary so we can receive the true connection state with subscription.


ERussel (commented):
Thanks! Could you, please, also fix it in the ```deactivateConnection``` function

svojsu:
> Thanks! Could you, please, also fix it in the `deactivateConnection` function

Yep, done

ERussel (approved):

PR #1109, opened 2024-06-24
Fix Ledger backup & more tests

ERussel:
## Purpose

- Ledger derivation path is converted to string before backup export and to chain codes data on backup import
- check that initial secrets in the Keychain matches one after sync
- add tests for backup wallets with chain accounts containing created based on mnemonic, seed and keystore


novawalletTests/Helper/KeystoreValidationHelper.swift
```diff
@@ -35,6 +35,36 @@ enum KeystoreValidationHelper {
...
+        for wallet: MetaAccountModel,
+        keystore: KeystoreProtocol
+    ) throws -> Bool {
+        let substrateTag = KeystoreTagV2.substrateDerivationTagForMetaId(wallet.metaId)
+        let ethereumTag = KeystoreTagV2.ethereumSeedTagForMetaId(wallet.metaId)
+        
+        let substrateDerPath = try keystore.loadIfKeyExists(substrateTag)
+        let ethereumDerPath = try keystore.loadIfKeyExists(ethereumTag)
+        
+        return substrateTag != nil && ethereumTag != nil
```
svojsu, line 65: why do we check tags if both `substrateDerPath` and `ethereumDerPath` didn't throw earlier?

svojsu (approved):

PR #1115, opened 2024-06-28
Feature/custom network add

svojsu:
Add/Edit feature for custom and known networks


novawallet/Common/Model/ChainRegistry/LocalChain/ChainModel.swift
```diff
@@ -533,3 +555,75 @@ extension ChainModel {
         return Set(priceIds)
     }
 }
+
+extension ChainModel {
+    static func partialCustomChainModel(
```
ERussel, line 560: Do we still need this if we are using separate model for partial chain?

novawallet/Modules/NetworkManagement/CustomNetwork/Presenter/CustomNetworkBasePresenter.swift
```diff
@@ -0,0 +1,272 @@
...
+
+extension CustomNetworkBasePresenter: CustomNetworkBaseInteractorOutputProtocol {
+    func didReceive(_ error: CustomNetworkBaseInteractorError) {
+        let close = R.string.localizable.commonClose(
+            preferredLanguages: selectedLocale.rLanguages
+        )
+        
+        let errorContent = error.toErrorContent(for: selectedLocale)
+        
+        wireframe.present(
```
ERussel, line 222: I think wireframe.present(error:...) is a shortcut for the error content presentable errors

ERussel (approved):

novawallet/Common/Model/ChainRegistry/LocalChain/ChainModel.swift
svojsu: No, we don't. Forgot to cut it out. Thanks!

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
```diff
@@ -0,0 +1,279 @@
...
+        )
+        
+        return wrapper
+    }
+}
+
+private extension CustomNetworkSetupTrait {
+    func createFillPartialChainWrapper(
+        partialChain: PartialCustomChainModel,
+        rawRuntimeFetchFactory: any RuntimeFetchOperationFactoryProtocol,
```
ERussel, line 72: Also, here why do we need to use any?

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
```diff
@@ -0,0 +1,279 @@
+import Foundation
+import SubstrateSdk
+import Operation_iOS
+
+protocol CustomNetworkSetupTrait: NetworkNodeCorrespondingTrait {
```
ERussel, line 5: I think a separate operation factory suits better here then trait as we create an operation that needs a lot of parameters. I would initialise the factory with parameters and have I single createWrapper() method

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
```diff
@@ -0,0 +1,279 @@
...
+        connection: ChainConnection
+    ) -> CompoundOperationWrapper<PartialCustomChainModel> {
+        let propertiesOperation: BaseOperation<SystemProperties>? = chain.isEthereumBased && chain.noSubstrateRuntime
+            ? nil
+            : systemPropertiesOperationFactory.createSystemPropertiesOperation(connection: connection)
+        
+        let assetOperation = ClosureOperation<PartialCustomChainModel> {
+            let properties = try propertiesOperation?.extractNoCancellableResultData()
+            let defaultEVMAssetPrecision: UInt16 = 18
+            let precision = properties?.tokenDecimals.first ?? defaultEVMAssetPrecision
```
ERussel, line 196: I would better separate the evm logic from substrate. Just to be sure we don't accedently set improper precision for substrate token

ERussel (approved):

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
svojsu: got rid of it

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
svojsu: done

novawallet/Modules/NetworkManagement/Traits/CustomNetworkSetupTrait.swift
svojsu: Agree, will do it as a separate PR

ERussel (approved):

PR #1123, opened 2024-07-05
filter out disabled chains

svojsu:


svojsu:
DAppBrowserInteractor also uses the chainRegistry to fetch all chains it operates with. Do we need to filter them too?

novawallet/Common/Services/ChainRegistry/ChainRegistry.swift
```diff
@@ -351,7 +351,7 @@ extension ChainRegistry: ChainRegistryProtocol {
     }
 
     func unsubscribeChainState(_ subscriber: ConnectionStateSubscription, chainId: ChainModel.Id) {
-        connectionPool.subscribe(subscriber, chainId: chainId)
+        connectionPool.unsubscribe(subscriber, chainId: chainId)
```
svojsu, line 354: btw seems like we there was a small mistake that prevented us from unsubscribing from the chain state

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+    private var transform: Transform? {
+        if case .enabledChains = self {
+            { change, currentChain in
+                guard let changedChain = change.item else { return change }
+
+                let currentSyncModeEnabled = currentChain.syncMode.enabled()
+                let updatedSyncModeEnabled = changedChain.syncMode.enabled()
+                let needsTransform = updatedSyncModeEnabled != currentSyncModeEnabled
+
+                return switch needsTransform {
```
ERussel, line 55: I think ```if ... then ... else if ...``` is more readable then this switch

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            nil
+        }
+    }
+    
+    func filter(
+        _ changes: [DataProviderChange<ChainModel>], 
+        using chainsBeforeChanges: [ChainModel.Id: ChainModel]
+    ) -> [DataProviderChange<ChainModel>] {
+        if case .noFilter = self {
```
ERussel, line 73: Didn't get why we need to handle this case separately

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+        let mapped = if let transform {
+            changes.map { change in
+                guard
+                    let changedChain = change.item,
+                    let currentChain = chainsBeforeChanges[changedChain.chainId]
+                else {
+                    return change
+                }
+
+                return transform(change, currentChain)
```
ERussel, line 86: I would do currentChain parameter optional and let transform function to decide what to do with it

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+        _ changes: [DataProviderChange<ChainModel>], 
+        using chainsBeforeChanges: [ChainModel.Id: ChainModel]
+    ) -> [DataProviderChange<ChainModel>] {
+        if case .noFilter = self {
+            return changes
+        }
+
+        let mapped = if let transform {
+            changes.map { change in
+                guard
```
ERussel, line 79: This guard doesn't make sense since we can get ```delete``` change that transform probably don't want to skip. Also, what we actually interested in is not an item but ```change.identifier```.

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+            let resultSet = Set(strategies.map { $0.filter(change) })
+            
+            return !resultSet.contains(false)
+        }
+        case .noFilter: { _ in true }
+        }
+    }
+
+    private var transform: Transform? {
+        if case .enabledChains = self {
```
ERussel, line 47: What if ```.enabledChains``` part of ```combined``` filter. I would handle that as well by passing the result of previous filter to the next one. I think we need this when we have, for example, ```.combined([.enabledChains, .hasProxy])```

novawallet/Common/Services/ChainRegistry/ChainRegistry.swift
ERussel: Yeah, my bad

novawallet/Common/Services/ChainRegistry/ChainRegistry.swift
```diff
@@ -293,11 +310,20 @@ extension ChainRegistry: ChainRegistryProtocol {
...
+        var currentChains = availableChains
+
         let updateClosure: ([DataProviderChange<ChainModel>]) -> Void = { changes in
             runningInQueue.async {
-                updateClosure(changes)
+                let filtered = filterStrategy.filter(
+                    changes,
+                    using: currentChains
+                )
+                currentChains = filtered.mergeToDict(currentChains)
```
ERussel, line 324: I think it will not work. Imagine a chain became disabled - then the filtered changes will contain .delete(...) change that will delete it from ```currentChains``` on merge. So, we need to merge not filtered but original changes I think. 

Only concern i have in this implementation is that we duplicate chains storage across all the subscriptions. I think better solution will be to store subscribers with filters directly in ChainRegistry and notify them in the end of the ```handle(changes:...)``` method. In this case we reuse existing availableChains storage for all subscribers and also guarantee that once they got chain runtimeProvider and connection is also in place to be queried - currently, we don't guarantee that.

novawallet/Common/Services/ExternalBalanceUpdater/CrowdloanService/CrowdloanOnChainSyncService.swift
```diff
@@ -119,6 +119,15 @@ final class CrowdloanOnChainSyncService: BaseSyncService {
     }
 
     override func performSyncUp() {
+        guard
```
ERussel, line 122: I think we don't need the logic here. Because we need it on the upper level to guarantee that we even not create the service if the chain is disabled.

novawallet/Modules/Vote/Crowdloan/CrowdloanList/CrowdloanListInteractor.swift
```diff
@@ -471,3 +476,16 @@ extension CrowdloanListInteractor: SelectedCurrencyDepending {
...
+extension CrowdloanListInteractor: EventVisitorProtocol {
+    func processNetworkEnableChanged(event: NetworkEnabledChanged) {
+        guard
+            let chain = crowdloanState.settings.value,
+            chain.chainId == event.chainId
+        else  {
+            return
+        }
+        
+        refresh(with: chain)
```
ERussel, line 489: The problem here is that CrowdloanChainSettings is not updating automatically. What we need to do is to call crowdloanState.settings.setup(...) and then refresh in callback with provided chain.

novawallet/Modules/Vote/Governance/Referendums/ReferendumsInteractor.swift
```diff
@@ -352,3 +357,11 @@ final class ReferendumsInteractor: AnyProviderAutoCleaning, AnyCancellableCleani
         operationQueue.addOperations(votingWrapper.allOperations, waitUntilFinished: false)
     }
 }
+
+extension ReferendumsInteractor: EventVisitorProtocol {
+    func processNetworkEnableChanged(event: NetworkEnabledChanged) {
+         if let option = governanceState.settings.value {
```
ERussel, line 363: The same is here. First, we need to check that the current option is changed chain. Then we need setup settings and get the new chain.

ERussel (commented):
Also, I don't see we have updated filters for ```CrowdloanChainSettings``` and ```GovernanceChainSettings```

novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
svojsu: This is to prevent unnecessary iterations with the functions below in case `.noFilter` is selected. This way, we can still call `filter(_ changes:, using:)` without worrying about it going through the changes if this is not needed.

novawallet/Common/Services/ChainRegistry/ChainRegistry.swift
```diff
@@ -123,6 +144,12 @@ final class ChainRegistry {
                 logger?.error("Unexpected error on handling chains update: \(error)")
             }
         }
+        
+        chainsChangesObservers.forEach { observer in
+            observer.queue.async {
```
ERussel, line 149: As I see in the chainsSubscribe closure itself already contains dispatch to the concrete queue

novawallet/Common/Services/ExternalBalanceUpdater/CrowdloanService/CrowdloanOffChainSyncService.swift
```diff
@@ -73,6 +73,11 @@ final class CrowdloanOffChainSyncService: BaseSyncService {
     }
 
     override func performSyncUp() {
+        guard chain.syncMode.enabled() else {
```
ERussel, line 76: Let's omit that, as we agreed it is not intended that the service will run for disabled chain. Also, instead of stopSyncUp()

ERussel (approved):

PR #1124, opened 2024-07-05
Fix/network management flow fixes

svojsu:
Added routings:

- Show manage custom network bottom sheet;
- Route to edit network;
- Show success alert on successful network add;
- Route back to networks list after successful network add;

Features:

- Skip remote nodes for networks if they were previously added as custom;
- Filter added networks in known networks list;
- Modify previously added network if user wants to add already existing custom network with different parameters;
- Switch to auto-balanced mode if user removes current selected custom node;

UI fixes:

- `chainUrlPlaceholder` changed to just a scheme;
- banner inner content spacings changed according to design;
- animations fixed to keep running after screen appear back in stack;
- Add Network button fix inactive state on appear back in stack;


novawallet/Modules/NetworkManagement/KnownNetworksList/KnownNetworksListViewController.swift
```diff
@@ -33,6 +33,11 @@ final class KnownNetworksListViewController: UIViewController, ViewHolder {
         super.viewDidLoad()
 
         setup()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
         presenter.setup()
```
ERussel, line 41: I would rename ```presenter.setup``` to ```presenter.becameActive``` to explicitly say that it might be called several times during lifecycle

novawallet/Modules/NetworkManagement/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -314,27 +307,65 @@ private extension NetworkDetailsInteractor {
...
+
+        if node == currentSelectedNode {
+            let chainWithAutoBalance = chain.updatingConnectionMode(for: .autoBalanced)
+
+            let setAutoBalanceOperation = repository.saveOperation(
+                { [chainWithAutoBalance] },
+                { [] }
+            )
+
+            let deleteNodeOperation = repository.saveOperation(
```
ERussel, line 349: Didn't get why do we need two operations, why just not have a single operation that saves the updated chain?

novawallet/Modules/NetworkManagement/NetworksList/NetworksListViewController.swift
```diff
@@ -104,21 +111,29 @@ extension NetworksListViewController: UITableViewDataSource {
             )
         }
     }
+
+    func tableView(_: UITableView, viewForFooterInSection _: Int) -> UIView? {
+        UIView()
```
ERussel, line 116: What we are trying to fix with that approach?

novawallet/Common/Services/ChainRegistry/ChainModelConversion.swift
```diff
@@ -34,12 +34,26 @@ final class ChainModelConverter: ChainModelConversionProtocol {
...
+
+            customNodes.append(node)
+            customNodesUrlSet.insert(node.url)
+        }
+
         let remoteNodes = remoteModel.nodes
-            .enumerated()
-            .map { ChainNodeModel(remoteModel: $1, order: Int16($0)) }
-        
+            .compactMap { node -> ChainNodeModel? in
```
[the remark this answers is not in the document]
svojsu: wanted to make it in single run instead of 2, but I agree, readability if much better with `.map` + `.filter`

novawallet/Modules/NetworkManagement/CustomNetwork/Interactor/CustomNetworkBaseInteractor.swift
```diff
@@ -39,15 +39,60 @@ class CustomNetworkBaseInteractor: NetworkNodeCreatorTrait,
...
+
+    func modify(
+        _ existingNetwork: ChainModel,
+        node: ChainNodeModel,
+        url: String,
+        name: String,
+        currencySymbol: String,
+        chainId: String?,
+        blockExplorerURL: String?,
+        coingeckoURL _: String?
```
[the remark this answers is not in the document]
svojsu: There is changes with coingecko url parsing and priceId extracting in the [following PR](https://github.com/novasamatech/nova-wallet-ios/pull/1126/commits/20fe5a6ffbdee04711ed595fd62c5a3988f626c1)

novawallet/Modules/NetworkManagement/CustomNetwork/Interactor/CustomNetworkBaseInteractor.swift
```diff
@@ -235,63 +280,66 @@ private extension CustomNetworkBaseInteractor {
...
-                
+
         return String(coingeckoUrlTemplate[matchedRange])
     }
-    
+
     func createExplorer(from url: String?) -> ChainModel.Explorer? {
-        guard let url = url?.trimmingCharacters(in: CharacterSet(charactersIn: "/")) else {
+        guard let url = url?.trimmingCharacters(
+            in: CharacterSet(charactersIn: "/").union(CharacterSet.whitespaces)
```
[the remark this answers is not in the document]
svojsu: `.slash` is a case of the `String.Separator` enum so it might be a bit misleading to use `String.Separator.slash.rawValue`. Also, we will have to wrap it with `String`, cause `.slash` has `Character` type and `CharacterSet` init needs `String` value. I find the result being too verbose in that case

novawallet/Modules/NetworkManagement/NetworksList/NetworksListViewController.swift
svojsu: It's just to create a space between sections according to the design, so we set a footer of the desired height for that.

novawallet/Modules/NetworkManagement/NetworkDetails/NetworkDetailsInteractor.swift
svojsu: There is some flaky behavior with single operation approach. Sometimes we don't receive `.connected` event on `WebSocketEngineDelegate` when updating connectionMode with removing node at the same time. 

Fixed, but further investigation and tests required.

novawallet/Modules/NetworkManagement/NetworkDetails/NetworkDetailsInteractor.swift
```diff
@@ -314,27 +307,65 @@ private extension NetworkDetailsInteractor {
...
+        _ node: ChainNodeModel,
+        chain: ChainModel
+    ) -> CompoundOperationWrapper<Void> {
+        let operation: (
+            targetOperation: BaseOperation<Void>,
+            dependencies: [BaseOperation<Void>]
+        )
+
+        if node == currentSelectedNode {
+//            let setAutoBalanceOperation = repository.saveOperation(
```
ERussel, line 342: Please, remove comments once you fix the issue

novawallet/Modules/NetworkManagement/KnownNetworksList/KnownNetworksListViewFactory.swift
```diff
@@ -2,29 +2,32 @@ import Foundation
 import SoraFoundation
 
 struct KnownNetworksListViewFactory {
-    static func createView() -> KnownNetworksListViewProtocol? {
+    static func createView(
+        successAddPresenting: (wireframe: ModalAlertPresenting, view: ControllerBackedProtocol)
```
ERussel, line 6: I thought we will improve the logic in all places not only in the CustomNetwork module - don't see a need to move ModalPresenting stuff around. Probably, let's do it in a separate PR.

ERussel (approved):

novawallet/Modules/NetworkManagement/KnownNetworksList/KnownNetworksListViewFactory.swift
svojsu: Damn, I was sure I removed it everywhere else. Thanks, I will remove it in separate PR for sure.

PR #1128, opened 2024-07-12
Fix/filter out disabled chains

svojsu:
- Filter enabled chains for some parts of the app;
- Remove `noFilter` for `ChainFilterStrategy`


novawallet/Common/Helpers/ChainsStore.swift
```diff
@@ -51,6 +61,10 @@ extension ChainsStore: ChainsStoreProtocol {
     }
 
     func getChain(for chainId: ChainModel.Id) -> ChainModel? {
-        chainRegistry.getChain(for: chainId)
+        let chain = chainRegistry.getChain(for: chainId)
```
ERussel, line 64: As we agreed let's also filter available chains to be consistent. So, I suggest filter availableChainIds applying filter to the corresponding chain (chainRegistry.getChain(...))

ERussel (commented):
Don't see where we filtered chains for the cross chain transfers

ERussel:
> Don't see where we filtered chains for the cross chain transfers

Nevermind, found changes

novawallet/Common/Helpers/ChainsStore.swift
svojsu: Thanks, done

ERussel (approved):

PR #1129, opened 2024-07-14
Assets balance single subscription

ERussel:
Instead of creating a separate subscription for each asset this PR introduces a service that allows to create single subscription for all assets per chain


svojsu (approved):

novawallet/Common/Services/TransactionSubscription/TransactionSubscription.swift
```diff
@@ -203,4 +122,124 @@ extension TransactionSubscription {
             }
         }
     }
+
+    private func createEventsWrapper(
+        depedingOn coderFactoryOperation: BaseOperation<RuntimeCoderFactoryProtocol>,
```
svojsu, line 127: a typo

PR #1130, opened 2024-07-15
UI fixes for network management

svojsu:
- fix placehoder for block explorer input;
- handle keyboard appearance with scroll on Custom Network addition screen;
- fix bottom network manage bottom sheet title;
- add SVG pencil icon instead of pdf so we can tint it everywhere we need to;


novawallet/Assets.xcassets/iconPencil.imageset/Contents.json
```diff
@@ -1,7 +1,7 @@
 {
   "images" : [
     {
-      "filename" : "iconPencil.pdf",
+      "filename" : "iconPencil.svg",
```
ERussel, line 4: Is there any pros we can get by using svg here instead of pdf?

ERussel (approved):

novawallet/Assets.xcassets/iconPencil.imageset/Contents.json
svojsu: The problem was not the format, but the icon that was opaque so it was impossible to tint it. Updated to new version of asset in pdf format.

PR #1131, opened 2024-07-15
Get rid of temporary fix for the TLS 1.3

ERussel:
- use TCPTransport instead of FoundationTransport to use native new web socket API and get TLS 1.3 support


novawallet/Common/Services/ChainRegistry/ConnectionPool/ConnectionFactory.swift
```diff
@@ -123,20 +119,18 @@ extension ConnectionFactory: ConnectionFactoryProtocol {
     }
 
     private func createConnection(
-        urlModels: [ConnectionTLSSupport],
+        urlModels _: [ConnectionCreationParams],
```
svojsu, line 122: Seems like we can remove it since no use. Or it's better to remove `urls` argument so we can use url from provided models.

svojsu (approved):

PR #1139, opened 2024-07-17
fix bugs for network management

svojsu:
- iOS. Not able to add network or node name with spaces;
- iOS. After switch to another node switch remains enabled;
- iOS. add custom EVM network: block exporer not displayed;
- search case sensitivity;
- add type for EVM asset on custom network setup;
- node ping operation added condition for noSubstrateRuntime;
- ChainFilterStrategy fix transforms;
- default explorer creation;

### DIFF
--- a/novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
+++ b/novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
@@ -90,13 +90,13 @@ enum ChainFilterStrategy {
         case let .chainId(chainId): { change, currentChain in
                 guard let changedChain = change.item else { return change }
 
-                let currentChainIdEquals = currentChain?.chainId == change.item?.chainId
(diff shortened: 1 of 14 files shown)


novawallet/Common/Services/ChainRegistry/ChainFilterStrategy.swift
```diff
@@ -25,14 +25,24 @@ enum ChainFilterStrategy {
...
                 #if F_RELEASE
                     return change.item?.hasProxy == true
                         && change.item?.isTestnet == false
                 #else
                     return change.item?.hasProxy == true
                 #endif
             }
-        case let .chainId(chainId): { $0.item?.chainId == chainId }
+        case let .chainId(chainId): { change in
+            if case .delete = change {
```
ERussel: I think we need to compare chainId with deleteIdentifier

svojsu: Right

PR #1142, opened 2024-07-17
add tests for chain filter strategy

svojsu:
added tests for ChainFilterStrategy

### DIFF
--- a/novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
+++ b/novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
@@ -1,0 +1,365 @@
+import XCTest
+import Operation_iOS
+@testable import novawallet
+
+class ChainFilterStrategyTests: XCTestCase {
+    
+    // MARK: hasProxy
+    
+    func testHasProxyFilterDelete() {
+        hasProxyFilterDeletes { updatedChains in
+            updatedChains.map { DataProviderChange<ChainModel>.update(newItem: $0) }
+        }
+        
+        hasProxyFilterDeletes { updatedChains in
+            updatedChains.map { DataProviderChange<ChainModel>.insert(newItem: $0) }
+        }
+    }
+    
+    func hasProxyFilterDeletes(changes: ([ChainModel]) -> [DataProviderChange<ChainModel>]) {
+        // given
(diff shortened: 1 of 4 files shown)


novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
```diff
@@ -0,0 +1,365 @@
...
+        let resultChanges = filterStrategy.filter(beforeFilterChanges, using: currentChains)
+        resultChains = resultChanges.mergeToDict(resultChains)
+        
+        // then
+        XCTAssert(updatedChains.allSatisfy { resultChains[$0.chainId] != nil })
+    }
+    
+    // MARK: enabledChains
+    
+    func testEnabledChainsFilterDelete() {
```
ERussel, line 89: Didn't get why it is called ```enabled chains``` test but we actually test proxy filter. Probably it is intended to call ```enabledChainsFilterDeletes``` not proxy one

novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
```diff
@@ -0,0 +1,365 @@
...
+        
+        // when
+        let resultChanges = filterStrategy.filter(beforeFilterChanges, using: [:])
+        resultChains = resultChanges.mergeToDict(resultChains)
+        
+        // then
+        XCTAssert(resultChains[chainId] != nil && resultChains.count == targetChainIdChains.count)
+    }
+    
+    func generateChainId(notMatching chainId: String) -> String {
```
ERussel, line 268: Don't think that this function makes sense in current implementation. There is very low possibility that chain ids will match given random function. So I would just go with ```Data.random```

ERussel (approved):

novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
svojsu: Tried to avoid any potential for test being flaky

novawalletTests/Common/Helpers/ChainFilterStrategyTests.swift
svojsu: Oh, thanks! Yes, you are right. A bit lack of attention on my side.

PR #1147, opened 2024-07-19
Make Nomination Pool row non tapable in staking details

svojsu:
don't show account options on selected pool tap


novawallet/Modules/Staking/StakingMain/StakingMainViewController.swift
```diff
@@ -160,7 +156,7 @@ final class StakingMainViewController: UIViewController, AdaptiveDesignable, Vie
 
             selectedEntityView = entityView
         }
-
+        entityView.
```
ERussel, line 159: Didn't get it. Typo?

svojsu: Oh, thanks. Not enough step-backs

ERussel (approved):

PR #1151, opened 2024-07-22
8.2.2 Security improvements

ERussel:
– Added support for new runtime upgrade affecting nomination pools
– Added security improvements for the DApp browser – Nova now rejects access to certain top level domains and warns users to be careful when connecting to DApps which are not already in the DApp catalog
– Fixes & Optimizations


ERussel:
Release severity: Normal
Release version: 8.2.2
Release notes:
Support for new runtime upgrades and security improvements are here

- Added support for new runtime upgrade affecting nomination pools
- Added security improvements for the DApp browser – Nova now rejects access to certain top level domains and warns users to be careful when connecting to DApps which are not already in the DApp catalog
- Fixes & Optimizations

svojsu (approved):

PR #1159, opened 2024-07-26
add localisation, removed unused strings

leohar:
Simulator Screenshot - iPhone 15 Pro - 2024-07-26 at 15 43 24


novawallet/en.lproj/Localizable.strings
```diff
@@ -521,8 +521,6 @@
 "wallet.balances.widget.title" = "Your balance";
 "common.max" = "max";
 "crowdloan.custom.contrib.format" = "%@ (via %@)";
-"username.setup.hint" = "";
-"username.setup.title" = "";
```
svojsu, line 525: I think we should remove it everywhere else, not just for english

svojsu (approved):

svojsu (approved):

PR #1165, opened 2024-07-28
Migrate to a long address prefix

svojsu:
- Migrate the `addressPrefix` field of `ChainModel` to `UInt64` so we can store longer `chainId` values for EVM networks.


ERussel (commented):
We need lightweight database migration to merge this PR, I think. But before doing this please make sure you have merged develop branch

ERussel (approved):

PR #1173, opened 2024-07-31
Feature/historical votes for ended referendums

svojsu:
- Display votes and progress for ended referendums
- Show skeleton view on loading


ERussel (commented):
We need also include delegatorVoting into off chain aye/nay votes so the votes on the details screen match voter screen

ERussel (approved):

PR #1175, opened 2024-08-01
fix abstain and standard voting responses mapping

svojsu:
- Fixed offchain abstain and standard voting results mapping
- Refactor to split the logic between Gov1 and OpenGov

### DIFF
--- a/novawallet/Modules/Vote/Governance/Operation/Fetch/OffchainVoting/GovernanceTotalVotesFactory.swift
+++ b/novawallet/Modules/Vote/Governance/Operation/Fetch/OffchainVoting/GovernanceTotalVotesFactory.swift
@@ -107,21 +107,35 @@ extension GovernanceTotalVotesFactory: GovernanceTotalVotesFactoryProtocol {
             prepareAllVotesQuery(referendumId: referendumId)
         }
 
-        let operation = createOperation(
-            for: query
-        ) { [weak self] (response: SubqueryVotingResponse.ReferendumVotesResponse) -> ReferendumVotingAmount in
-            response.castingVotings.nodes
-                .compactMap { ReferendumVoterLocal(from: $0) }
-                .reduce(
-                    ReferendumVotingAmount(
-                        aye: 0,
-                        nay: 0,
-                        abstain: 0
-                    )
-                ) { self?.sum($0, with: $1) ?? $0 }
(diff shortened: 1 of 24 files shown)


novawallet/Modules/Vote/Governance/ViewModel/ReferendumDisplayStringFactory.swift
```diff
@@ -12,13 +12,135 @@ protocol ReferendumDisplayStringFactoryProtocol {
...
+        let loadableValueString: LoadableViewModelState<String>
+
+        if
+            let value,
+            let valueString = createVotes(
+                from: value,
+                chain: chain,
+                locale: locale
+            ) {
+            loadableValueString = .loaded(value: valueString)
```
ERussel, line 76: I don't like that approach. The factory intended to be one that returns formatted string. It shouldn't decide whether to ask for loading - it obviously don't have enough information. For abstain we already have supportsAbstain flag passed to the presenter so we can decide based on this flag what to display. For displaying off chain votes we also can introduce ```supportsOffchainVotes``` flag to make the behaviour simpler and obvious.

ERussel (approved):

novawallet/Modules/Vote/Governance/Operation/Fetch/OffchainVoting/GovernanceTotalVotesFactory.swift
```diff
@@ -107,23 +107,37 @@ extension GovernanceTotalVotesFactory: GovernanceTotalVotesFactoryProtocol {
...
+        let operation = if votersType == .abstains {
+            createOperation(for: query, resultHandler: mapAbstainVotingResponse)
+        } else {
+            createOperation(for: query, resultHandler: mapVotingResponse)
         }
 
         return operation
     }
 
+    private func mapAbstainVotingResponse(_ response: SubqueryVotingResponse.CastingResponse) -> ReferendumVotingAmount {
```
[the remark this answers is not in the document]
svojsu: fixed

novawallet/Modules/Vote/Governance/ViewModel/ReferendumDisplayStringFactory.swift
svojsu: fixed

PR #1176, opened 2024-08-02
fix referendum curves and prediction logic

svojsu:
- make decimals same scale to compare
- add rounding for curves logic
- fix threshold calculations for linear stepped curve


novawallet/Common/Extension/Foundation/Decimal+Conversion.swift
```diff
@@ -66,4 +66,74 @@ extension Decimal {
 
         return decimal2 / decimal1
     }
+
+    func greaterThanOrEqual(
```
ERussel, line 70: Why it is important to have that functions?

svojsu: The scale of compared values may differ. For example, if we compare 0.957449566 to 0.9575, we may need to round them to the same scale. A good example is referendum approval logic, where we compare fraction values to a more precise threshold value. The threshold value may appear the same when rounded to a lower scale but can be different when compared at a higher scale.

ERussel (approved):

PR #1179, opened 2024-08-05
Fix/remove filter for notifications

svojsu:
- Show notifications for disabled networks.
- Display disabled networks in notification settings.


novawallet/Modules/Notifications/StakingRewardsNotifications/StakingRewardsNotificationsInteractor.swift
```diff
@@ -13,8 +13,7 @@ final class StakingRewardsNotificationsInteractor {
     private func subscribeChains() {
         chainRegistry.chainsSubscribe(
             self,
-            runningInQueue: .main,
-            filterStrategy: .enabledChains
```
ERussel, line 17: I think we need also remove filters for governance push notifications

svojsu: we don't use filters for governance notifications

ERussel (approved):

PR #1181, opened 2024-08-06
Feature/applinks

svojsu:
- Add support for app links
- Process app links the same way as deeplinks


novawallet/Common/Configs/ApplicationConfigs.swift
```diff
@@ -301,5 +301,13 @@ extension ApplicationConfig: ApplicationConfigProtocol {
         URL(string: "https://docs.novawallet.io/nova-wallet-wiki/wallet-management/delegated-authorities-proxies")!
     }
 
+    var novaWalletURL: URL {
```
ERussel, line 304: Too broad name I think. Let's make it more concrete. ```universalLinkUL```?

novawallet/Common/URLHandling/UrlHandlingAction.swift
```diff
@@ -1,6 +1,25 @@
...
-enum UrlHandlingAction: String {
-    case open
-    case create
+enum UrlHandlingAction {
+    case open(screen: String)
+    case create(screen: String)
+
+    init?(from url: URL) {
+        let pathComponents = url.pathComponents
+        let action = pathComponents[1].lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
```
ERussel, line 9: I think we should check whether path components at a given index exists. Otherwise we will crash.

ERussel (approved):

PR #1187, opened 2024-08-13
Feature/asset hub custom fee

svojsu:
Add a feature to change the fee asset for Polkadot Asset Hub transfers.


novawallet/Modules/Transfer/BaseTransfer/OnChain/OnChainTransferPresenter.swift
```diff
@@ -113,18 +123,18 @@ class OnChainTransferPresenter {
             ),
 
             dataValidatingFactory.canPayFeeSpendingAmountInPlank(
-                balance: senderUtilityAssetTransferable,
+                balance: senderAssetTransferable,
                 fee: fee?.value,
-                spendingAmount: isUtilityTransfer ? sendingAmount : nil,
-                asset: utilityAssetInfo,
+                spendingAmount: feeAsset == chainAsset ? sendingAmount : nil,
```
ERussel, line 128: it is safer to compare chainAssetId as it never changes

ERussel (commented):
What I thought about is that we probably need an API to check if asset is supported for fee

novawallet/Modules/Transfer/BaseTransfer/OnChain/OnChainTransferPresenter.swift
svojsu: Yep, updated

novawallet/Modules/Transfer/TransferSetup/FeeAssetSelectionPresentable.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        view?.controller.present(bottomSheet.controller, animated: true)
+    }
+
+    private func createViewModel(
+        utilityAsset: ChainAsset,
+        sendingAsset: ChainAsset,
+        currentFeeAsset: ChainAsset?,
+        onFeeAssetSelect: ((ChainAsset) -> Void)?
+    ) -> SwapNetworkFeeSheetViewModel {
```
ERussel, line 45: Look like this bottom sheet became more general. Let's rename to something like EditNetworkFeeSheet

novawallet/Modules/Transfer/TransferSetup/TransferSetupPresenter.swift
```diff
@@ -35,6 +36,8 @@ final class TransferSetupPresenter {
         }
     }
 
+    private var isManualFeeSet: Bool = false
```
ERussel, line 39: What is this guy for?

ERussel (commented):
Overall looks good. But I didn't find logic that automatically switches to custom fee when no utility token on user's account. That's why I guess we have ```isManualFeeSet``` flag unused.

novawallet/Modules/Transfer/TransferSetup/TransferSetupPresenter.swift
svojsu: Forgot to remove it since it became redundant at the end. Thanks!

novawallet/Modules/Transfer/BaseTransfer/OnChain/EvmOnChainTransferInteractor.swift
```diff
@@ -167,6 +167,10 @@ extension EvmOnChainTransferInteractor {
         }
     }
 
+    func requestFeePaymentAvailability(for _: ChainAsset) {
+        presenter?.didReceiveSendingAssetFeeAvailable(false)
```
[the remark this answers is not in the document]
svojsu: yeah, confusing naming happened

novawallet/Modules/Transfer/BaseTransfer/OnChain/OnChainTransferInteractor.swift
```diff
@@ -535,6 +547,26 @@ extension OnChainTransferInteractor {
...
+        execute(
+            wrapper: wrapper,
+            inOperationQueue: operationQueue,
+            runningCallbackIn: .main,
+            callbackClosure: { [weak self] result in
+                switch result {
+                case let .success(available):
+                    self?.presenter?.didReceiveSendingAssetFeeAvailable(available)
+                case let .failure(error) where error is AssetConversionAggregationFactoryError:
+                    self?.presenter?.didReceiveSendingAssetFeeAvailable(false)
```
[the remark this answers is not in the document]
svojsu: The wrapper factory encapsulates the logic for checking whether `hasAssetHubTransferFees` or `hasHydrationTransferFees` is available. If neither is available, it wraps the error. I believe this approach is better since it's a common logic, so we don't need to implement it at the upper level, such as in interactors.

novawallet/Modules/Transfer/Operation/AssetTransferAggregationFactory.swift
```diff
@@ -59,9 +59,9 @@ final class AssetTransferAggregationFactory: AssetTransferAggregationFactoryProt
     }
 
     func createCanPayFeeWrapper(in chainAsset: ChainAsset) -> CompoundOperationWrapper<Bool> {
-        if chainAsset.chain.hasSwapHub {
+        if chainAsset.chain.hasAssetHubTransferFees {
             return createAssetHubCanPayFee(for: chainAsset)
-        } else if chainAsset.chain.hasSwapHydra {
+        } else if chainAsset.chain.hasHydationTransferFees {
```
svojsu, line 64: These changes were made previously but were lost during a bit of code restructuring. Fortunately, the manually tested chain has both `hasSwapHub` and `hasAssetHubTransferFees`, so it was still valid. However, I retested it to ensure that it works as before

ERussel (approved):

PR #1189, opened 2024-08-18
Show fee in a custom asset for OperationDetails module

svojsu:
- Parse the transfer extrinsic to find the 'AssetTxPaymentPallet.assetTxFeePaidEvent' event
- Use the 'assetId' from the event as the fee asset
- Adapt the 'OperationDetails' module to show the fee in a custom asset

### DIFF
--- a/novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Fee.swift
+++ b/novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Fee.swift
@@ -3,14 +3,26 @@ import BigInt
 import SubstrateSdk
 
 extension ExtrinsicProcessor {
+    struct Fee {
+        let amount: BigUInt
+        let assetId: JSON?
+    }
+
     func findFee(
         for index: UInt32,
         sender: AccountId,
         eventRecords: [EventRecord],
         metadata: RuntimeMetadataProtocol,
         runtimeJsonContext: RuntimeJsonContext
-    ) -> BigUInt? {
-        if let fee = findTransactionFeePaid(
+    ) -> Fee? {
+        if let fee = findTransactionFeePaidInCustomAsset(
+            for: index,
+            eventRecords: eventRecords,
+            metadata: metadata,
+            runtimeJsonContext: runtimeJsonContext
+        ) {
+            return fee
+        } else if let fee = findTransactionFeePaid(
             for: index,
             eventRecords: eventRecords,
             metadata: metadata,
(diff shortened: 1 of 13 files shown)


novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Fee.swift
```diff
@@ -35,25 +47,54 @@ extension ExtrinsicProcessor {
         }
     }
 
+    private func findTransactionFeePaidInCustomAsset(
```
ERussel, line 50: I think we can reuse existing function ExtrinsicProcessor.findFeeInCustomAsset

novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Matching.swift
```diff
@@ -665,4 +675,25 @@ extension ExtrinsicProcessor {
             return nil
         }
     }
+
+    private func localAsset(
```
ERussel, line 679: method name look to general. Probably we can move it to AssetHubTokensConverter

novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Matching.swift
```diff
@@ -90,6 +90,11 @@ extension ExtrinsicProcessor {
                 runtimeJsonContext: context
             )
 
+            let feeAsset = localAsset(
```
ERussel, line 93: Here we need to have orml specific parsing. More precisely the parser must match the chain and apply Hydra specific parsing.

ERussel (commented):
Overall I wouldn't merge it now since we don't have subquery ready for the custom fees.

novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Fee.swift
```diff
@@ -35,25 +40,85 @@ extension ExtrinsicProcessor {
...
+            guard
+                let feeAssetId = chain.utilityChainAssetId()?.assetId,
+                let nativeFee = optNativeFee?.amount
+            else {
+                return nil
+            }
+
+            return Fee(
+                amount: nativeFee,
+                assetId: feeAssetId
```
ERussel, line 94: In other place, I saw we returning nil for native fee

ERussel (approved):

novawallet/Common/Services/TransactionSubscription/ExtrinsicProcessor+Fee.swift
svojsu: Oh, I will check before the merge. Thanks.

PR #1190, opened 2024-08-19
use new extrinsic logic for hydration swaps

svojsu:
Use the new extrinsic service logic for the hydration swap flow


ERussel (commented):
One thing I noticed: currently we are determining whether to add set currency call based on provided asset. It is not the best approach because we are missing corner case when a user has custom fee already set but we are assuming it is HDX. We need to use swap params to determine that. But I would do it in a separate PR.

ERussel (approved):

PR #1202, opened 2024-08-29
Feature/External api support

svojsu:
- Parse and utilize the 'referendumSummary' external API.
- Display the TinderGov banner based on support provided by the 'referendumSummary' API.
- Perform a simple code refactoring to improve the readability and maintainability of the 'ReferendumsPresenter'.


novawallet/Modules/Vote/Governance/TinderGov/TinderGovViewModelFactory.swift
```diff
@@ -3,20 +3,24 @@ import Foundation
...
+        accountVotes: ReferendumAccountVotingDistribution?,
         locale: Locale
     ) -> ReferendumsSection? {
         guard let referendums else {
             return nil
         }
 
-        let tinderGovReferenda = referendums.filter { !$0.state.completed && $0.canVote }
+        let tinderGovReferenda = referendums.filter {
+            $0.canVote && (accountVotes?.votes[$0.index] == nil)
```
ERussel, line 22: One also can't vote for a referendum if the referendum is in the track in which one is delegating. So, we also need to check ```accountVotes.delegatings```

ERussel (commented):
Also I wonder how we gonna the reuse the filter for tindergov as it is currently defined in the view model factory. Probably we need introduce a separate struct/class that defines that filter to reuse in the view model factory and in the tindergov logic

svojsu:
> Also I wonder how we gonna the reuse the filter for tindergov as it is currently defined in the view model factory. Probably we need introduce a separate struct/class that defines that filter to reuse in the view model factory and in the tindergov logic

will present in the next PR

ERussel (approved):

PR #1205, opened 2024-09-06
Feature/tindergov cards screen logic

svojsu:
- The governance shared state now includes an observable referendums store.
- `VoteAvailableReferendumSource` wraps `ReferendumObservableSource` and filters changes based on the available referendums logic tailored to TinderGov's needs.
- The new referendum list matches the visible cards.
- Content is loading for the top 3 cards.
- Skip and retry options are available for the top card if content loading fails.


novawallet/Modules/Vote/Governance/TinderGov/Helpers/ReferendumObservableSource.swift
```diff
@@ -0,0 +1,128 @@
...
+    var observers: [WeakWrapper] = []
+    private var referendums: [ReferendumIdLocal: ReferendumLocal] = [:]
+
+    private let mutex = NSLock()
+}
+
+// MARK: ReferendumObservableSourceProtocol
+
+extension ReferendumObservableSource: ReferendumObservableStoreProtocol, WeakWrappersClearing {
+    func save(referendums: [ReferendumLocal]) {
```
ERussel, line 37: We call save in the ReferendumsInteractor and never call remove. So will have still have removed referendums in the list.

novawallet/Modules/Vote/Governance/TinderGov/TinderGovPresenter.swift
```diff
@@ -39,20 +49,50 @@ extension TinderGovPresenter: TinderGovPresenterProtocol {
 // MARK: TinderGovInteractorOutputProtocol
 
 extension TinderGovPresenter: TinderGovInteractorOutputProtocol {
-    func didReceive(_ referendums: [ReferendumLocal]) {
-        self.referendums = referendums
+    func didReceive(_ changes: [DataProviderChange<ReferendumLocal>]) {
+        referendums = changes.mergeToDict(referendums)
+
+        sortedReferendums = referendums.values.sorted {
```
ERussel, line 55: I would move that logic into the separate class. Check ```AssetListBuilder``` for reference.

novawallet/Modules/Vote/Governance/TinderGov/Helpers/VoteAvailableReferendumSource.swift
```diff
@@ -0,0 +1,62 @@
...
+        defer { mutex.unlock() }
+
+        clearEmptyWrappers()
+
+        let weakWrapper = WeakWrapper(target: target)
+        observers.append(weakWrapper)
+
+        guard !changes.isEmpty else { return }
+
+        target.didReceive(Array(changes.values))
```
ERussel, line 39: So, here might be updates and we consider only inserts in the TinderGovPresenter

ERussel (commented):
Current approach with sharing referendums state requires a pipeline of filters that is complicated (I highlighted potential bugs not even running the code) and as result determines the actual changes not precisely. Also adding account votes to the state further complicates that approach.

I would still go with the obvervable model approach that has been tested across the app. For example, in the AssetsList. And significantly simpler. The problem with equitable can be solved with ```NotEqualWrapper``` as in the ```AssetListModelObservable```.  It also not worth to add TinderGovModelBuilder to the interactor to calculate specific model for the TinderGov as we are doing in the AssetsListBuilder instead of doing it in the main thread in the presenter.

novawallet/Modules/Vote/Governance/Referendums/Presenter/ReferendumsPresenter.swift
```diff
@@ -35,13 +35,19 @@ final class ReferendumsPresenter {
...
         didSet {
-            observableState.state.timeModels = timeModels
+            observableViewState.state.timeModels = timeModels
         }
     }
 
     private(set) var filter = ReferendumsFilter.all
 
-    let observableState = Observable<ReferendumsState>(state: .init(cells: [], timeModels: nil))
+    let tinderGovObservableState = Observable<NotEqualWrapper<[ReferendumIdLocal: ReferendumLocal]>>(
```
ERussel, line 44: I would make it generic and move tinder gov specific logic (filtering/sorting) to TinderGovBuilder.

novawallet/Modules/Vote/Governance/Referendums/Model/ReferendumsViewState.swift
```diff
@@ -0,0 +1,8 @@
+struct ReferendumsViewState: Equatable {
+    var cells: [ReferendumsCellViewModel]
+    var timeModels: [ReferendumIdLocal: StatusTimeViewModel?]?
+}
+
+struct ReferendumsState {
+    var referendums: [ReferendumLocal]
```
ERussel, line 7: What about account voting? And Why it is var and not let?

novawallet/Modules/Vote/Governance/TinderGov/Helpers/DataProviderChange+ReferendumLocal.swift
```diff
@@ -0,0 +1,30 @@
...
+            return newItem.index
+        case let .update(newItem):
+            return newItem.index
+        case let .delete(deletedIdentifier):
+            return UInt(deletedIdentifier)!
+        }
+    }
+}
+
+extension Array where Element == DataProviderChange<ReferendumLocal> {
```
ERussel, line 17: I don't think we need these extensions now as we are working with another model

novawallet/Modules/Vote/Governance/TinderGov/TinderGovViewFactory.swift
```diff
@@ -2,17 +2,57 @@ import Foundation
 import SoraFoundation
 
 struct TinderGovViewFactory {
-    static func createView(with referendums: [ReferendumLocal]) -> TinderGovViewProtocol? {
+    static func createView(
+        observableState: Observable<NotEqualWrapper<[ReferendumIdLocal: ReferendumLocal]>>,
+        sharedState: GovernanceSharedState
```
ERussel, line 7: As we agreed let's put observable state to the shared state and clean on account/chain change

novawallet/Modules/Vote/Governance/TinderGov/Views/CardsZStack/VoteCardView.swift
```diff
@@ -46,7 +48,25 @@ final class VoteCardView: RoundedView {
         )
     }
 
-    private var viewModel: VoteCardViewModel?
+    private(set) var viewModel: VoteCardViewModel?
+
+    var isLoading: Bool = false {
+        didSet {
+            isSummaryLoading = isLoading
```
ERussel, line 55: That one looks strange: didStart/Stop loading changes both summary loading and isRequestedAmountLoading. As we want 2 parts to be independent then I would move them to the separate view and both views to the card. Also if we separate views we can also separate view model. It will reduce complexity that we talked about big single view model

novawallet/Modules/Vote/Governance/TinderGov/TinderGovViewFactory.swift
svojsu: Will present in the coming PR

novawallet/Modules/Vote/Governance/Referendums/Model/ReferendumsViewState.swift
svojsu: The model is not in use. There were plans for it, but not anymore.

novawallet/Modules/Vote/Governance/Referendums/Presenter/ReferendumsPresenter.swift
svojsu: Will present in the coming PR

ERussel (approved):

PR #1225, opened 2024-10-07
Nova Card opening

ERussel:
## Purpose

- introduce ```PayCardHook``` notion to inject scripts into WKWebView and handle events
- introduce ```PayCardHookFactoryProtocol``` and ```MercuryoCardHookFactory```
- introduce Mercuryo handlers fro cards response, on status change and on topup
- move entrypoint to Assets

##

IMG_B09753B42982-1


novawallet/Modules/PayCard/PayCardInteractor.swift
```diff
@@ -3,46 +3,82 @@ import Foundation
 final class PayCardInteractor {
     weak var presenter: PayCardInteractorOutputProtocol?
 
-    let chainRegistry: ChainRegistryProtocol
+    let payCardModelFactory: PayCardHookFactoryProtocol
```
svojsu, line 6: It seems that `payCardModelFactory` is a bit misleading. `PayCardModel` is created below without using any model factory, and `payCardModelFactory` is actually a hooks factory. `payCardHookFactory` might be a better fit here.

svojsu (approved):

PR #1229, opened 2024-10-08
Feature/Card pending sheet

svojsu:
- Show bottomsheet with timer when card is not ready yet;
- Fix naming;
- Fix widget margin resulted in white border;


novawallet/Modules/PayCard/PayCardPresenter.swift
```diff
@@ -1,16 +1,75 @@
...
+    }
+
+    private func showCardOpenPending() {
+        guard let openCardTimestamp else {
+            return
+        }
+
+        let timeRemained = Date().timeIntervalSince1970 - openCardTimestamp
+        
+        guard timeRemained < 60*5 else {
```
ERussel, line 58: Let's move the time to the constants

novawallet/Modules/PayCard/PayCardPresenter.swift
```diff
@@ -1,16 +1,75 @@
...
+    private func showFailAlert() {
+        let languages = localizationManager.selectedLocale.rLanguages
+
+        let title = R.string.localizable.cardOpenFailedAlertTitle(preferredLanguages: languages)
+        let message = R.string.localizable.cardOpenFailedAlertMessage(preferredLanguages: languages)
+
+        let closeAction = AlertPresentableAction(
+            title: R.string.localizable.commonOk(preferredLanguages: languages),
+            handler: { [weak self] in
+                self?.wireframe.close(from: self?.view)
```
ERussel, line 33: Why do we close the widget? Probably user want's to retry or get any information from the state

novawallet/en.lproj/Localizable.strings
```diff
@@ -1762,3 +1762,8 @@
 "common.use.max.due.fee.message" = "You can use up to %@ since you need to pay %@ for network fee.";
 "common.pay.anywhere" = "Pay anywhere";
 "common.nova.card" = "Nova Card";
+"common.estimated.timer" = "Estimated ~%@";
```
ERussel, line 1765: We have ```approximately()``` wrapper in ```String``` extension. I would better use it

novawallet/en.lproj/Localizable.strings
```diff
@@ -1762,3 +1762,8 @@
...
+"card.open.pending.sheet.title" = "Your card is being issued!";
+"card.open.pending.sheet.message" = "It can take up to 5 minutes.\nThis window will be closed automatically.";
```
ERussel, line 1767: Let's parametrise by the value of time we have in ```PayCardPresenter``` instead of hardcoding 5 minutes

novawallet/Modules/PayCard/PayCardInteractor.swift
```diff
@@ -61,6 +73,19 @@ extension PayCardInteractor: PayCardInteractorInputProtocol {
...
+        guard settingsManager.novaCardStatus != .created else {
+            return
+        }
+
+        let pendingStatus = PayCardStatus.pending
+
+        settingsManager.novaCardOpenTimeStamp = UInt64(Date().timeIntervalSince1970)
+        settingsManager.novaCardStatus = pendingStatus
+
+        presenter?.didReceiveCardStatus(pendingStatus)
```
ERussel, line 86: We don't send timestamp here to the presenter after topup and that is the reason why the pending bottom sheet is not shown

novawallet/Modules/PayCard/PayCardInteractor.swift
```diff
@@ -38,10 +42,18 @@ final class PayCardInteractor {
 
 extension PayCardInteractor: PayCardInteractorInputProtocol {
     func setup() {
+        if let cardOpenStartTimestamp = settingsManager.novaCardOpenTimeStamp {
+            presenter?.didReceiveCardOpenTimestamp(cardOpenStartTimestamp.timeInterval)
```
ERussel, line 46: ```.timestamp``` assumes value is in milliseconds. It should be TimeInterval(cardOpenStartTimestamp) I think

ERussel:
Will open a separate one

PR #1233, opened 2024-10-09
Fix swipe gov proxy support

ERussel:
Take into account ```fee.currentAccountAmount``` maybe nil when voting from proxy. In general we just reuse validations from regular gov voting that already takes proxy into account

### DIFF
--- a/novawallet/Modules/Vote/Governance/Validating/GovernanceValidatorFactory.swift
+++ b/novawallet/Modules/Vote/Governance/Validating/GovernanceValidatorFactory.swift
@@ -7,6 +7,7 @@ protocol GovernanceValidatorFactoryProtocol: BaseDataValidatingFactoryProtocol {
         _ assetBalance: AssetBalance?,
         votingAmount: BigUInt?,
         assetInfo: AssetBalanceDisplayInfo,
+        maxAmountErrorClosure: ((BigUInt) -> Void)?,
         locale: Locale?
     ) -> DataValidating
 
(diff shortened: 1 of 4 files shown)


novawallet/Modules/Vote/Governance/Validating/GovernanceValidatorFactory.swift
```diff
@@ -166,31 +97,38 @@ extension GovernanceValidatorFactory: GovernanceValidatorFactoryProtocol {
...
-            )
         }, preservesCondition: {
             guard
                 let assetBalance = assetBalance,
                 let votingAmount = votingAmount else {
                 return false
             }
 
-            return assetBalance.freeInPlank >= votingAmount
+            return assetBalance.availableForOpenGov >= votingAmount
```
svojsu, line 131: `availableForOpenGov` validating `votingAmount` for common gov validation needs looks a bit misleading. Possibly, we should use `freeInPlank` here.

ERussel: Why it is misleading? We are using ```availableForOpenGov``` abstraction to check voting amount in the ```enoughTokenForVoting``` method. I think I have moved to that abstraction to most places in swipegov and it is still uses ```freeInPlank``` under the hood. We have another task to complete migration in all places in the opengov and change the logic behind ```availableForOpenGov```

ERussel:
Decided to go via hotfix to main

PR #1236, opened 2024-10-10
Fix swipe gov proxy support

ERussel:
Take into account ```fee.currentAccountAmount``` maybe nil when voting from proxy. In general we just reuse validations from regular gov voting that already takes proxy into account


svojsu (approved):

ERussel:
Release severity: Normal
Release version: 8.7.1
Release notes:
8.7.1 SwipeGov & Governance Improvements
- Add proxy support for swipegov

PR #1252, opened 2024-10-23
Feature/move zero balances filter

svojsu:
- Move "hidesZeroBalances" filter to `TokensManage` module
- Remove settings button from `AssetList` module
- Fix tests to check filter event is sent through the event center
- Remove `AssetSettings` module


novawallet/Common/View/TopCustomSearchView.swift
```diff
@@ -1,7 +1,7 @@
 import UIKit
 
-final class TopCustomSearchView: UIView {
-    static let preferredBarHeight: CGFloat = 48.0
+class TopCustomSearchView: UIView {
+    static let preferredBarHeight: CGFloat = 98.0
```
ERussel, line 4: I see that this value only used in the TokensManage I would just move it there to the constants otherwise it is confusing that we are changing it here but the view itself is used in many different places

ERussel (approved):

PR #1272, opened 2024-11-10
Fix/balance on network selection

svojsu:
- pass localization manager to `AssetOperationNetworkListPresenter`
- pass selected locale to `AssetListCollectionManager` on localization change
- show transferrable balance for send and swaps operations


novawallet/Modules/AssetsSearch/AssetOperation/Swaps/SwapAssetsOperationInteractor.swift
```diff
@@ -13,7 +13,7 @@ final class SwapAssetsOperationInteractor: AnyCancellableCleaning {
     let settingsManager: SettingsManagerProtocol
 
     private let operationQueue: OperationQueue
-    private var builder: AssetSearchBuilder?
+    private var builder: SendAssetSearchBuilder?
```
ERussel, line 16: It is also model named as send passed to swaps. Let's also rename it to SpendAssetSearchBuilder

ERussel (approved):

PR #1273, opened 2024-11-11
DApp Browser tabs state managing PoC

svojsu:
Use the `Add` and `Nova` buttons at the bottom to add new tabs and manage the tab list.

### DIFF
--- a/novawallet/Common/Extension/SettingsExtension.swift
+++ b/novawallet/Common/Extension/SettingsExtension.swift
@@ -25,6 +25,7 @@ enum SettingsKey: String {
     case integrateNetworksBannerSeen
     case assetListGroupStyle
     case assetIconsAppearance
+    case webViewStates
 }
 
 extension SettingsManagerProtocol {
@@ -292,6 +293,17 @@ extension SettingsManagerProtocol {
                 value: newValue.rawValue,
                 for: SettingsKey.assetIconsAppearance.rawValue
             )
+        }
+    }
+
+    var webViewStates: [String: Data]? {
+        get {
+            let anyValue = anyValue(for: SettingsKey.webViewStates.rawValue)
+            return anyValue as? [String: Data]
+        }
+
+        set {
+            set(anyValue: newValue, for: SettingsKey.webViewStates.rawValue)
         }
     }
 }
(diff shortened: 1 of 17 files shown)


novawallet/Common/Extension/SettingsExtension.swift
```diff
@@ -294,4 +295,15 @@ extension SettingsManagerProtocol {
             )
         }
     }
+
+    var webViewStates: [String: Data]? {
```
ERussel, line 299: Not a good place for this. I would consider single value in the database or file

novawallet/Modules/DApp/DAppBrowser/DAppBrowserPresenter.swift
```diff
@@ -51,6 +56,20 @@ final class DAppBrowserPresenter {
 
         view?.didSet(canShowSettings: canShowSettings)
     }
+
+    private func saveCurrentWebViewState() {
+        if
+            let currentTabId,
+            let currentWebView = webViewPool.getWebView(for: currentTabId),
```
ERussel, line 63: I would probably pass pool directly to the controller via init. And controller can call presenter to save opaque state given in the parameters

novawallet/Modules/DApp/DAppBrowser/DAppBrowserPresenter.swift
```diff
@@ -98,6 +117,33 @@ extension DAppBrowserPresenter: DAppBrowserPresenterProtocol {
         )
     }
 
+    func showTabs() {
+        wireframe.presentTabs(
+            from: view
+        ) { [weak self] selectedId in
```
ERussel, line 123: What is about to return tab directly?

novawallet/Modules/DApp/DAppBrowser/DAppBrowserTabs/DAppBrowserTabsViewController.swift
```diff
@@ -0,0 +1,93 @@
+import UIKit
+
+typealias DappBrowserTabCell = PlainBaseTableViewCell<UILabel>
```
ERussel, line 3: I saw that on the PR screenshots there is an image per cell but can't find it. Probably forget to push

svojsu: The second screenshot shows dotappstore, which appears to be tabs in a screenshot format, but it isn't. I haven't yet implemented screenshots or mature persistence using CoreData, as I want to avoid dealing with migrations later in this branch. The tab model should be properly planned first.

novawallet/Common/Extension/SettingsExtension.swift
svojsu: Yeah, sure. It was intended just to PoC so we can persist webview's state and restore it then. I haven't yet implemented persistence using CoreData, as I want to avoid dealing with migrations later in this branch. The tab model should be properly planned first.

PR #1287, opened 2024-11-21
Feature/dapp browser tabs

svojsu:
- Introduce `DAppBrowserTabManager` to manage persistence and in-memory state for tabs.
- Introduce `WebViewPool` to reuse webviews.
- Introduce `WebViewRenderFilesOperationFactory` to manage cache for webview renders.
- Introduce `DAppBrowserTabList` module
- Refactor `DAppBrowserInteractor` to deal with tabs instead of dApps or just user queries.


novawallet/Modules/DApp/DAppBrowser/Tabs/DAppBrowserTabManager/DAppBrowserTabManager.swift
```diff
@@ -0,0 +1,368 @@
...
+    private let fileRepository: WebViewRenderFilesOperationFactoryProtocol
+    private let repository: CoreDataRepository<DAppBrowserTab.PersistenceModel, CDDAppBrowserTab>
+    private let operationQueue: OperationQueue
+    private let observerQueue: DispatchQueue
+
+    private let logger: LoggerProtocol
+
+    private var dAppTransportStates: [UUID: [DAppTransportState]] = [:]
+    private var tabs: [UUID: DAppBrowserTab] = [:]
+    private var observers: [WeakWrapper] = []
```
ERussel, line 15: I think we completely forgot here about primitives we have to organise reactive code based on the database. I would suggest the following approach:

1) Make `PersistentTabLocalSubscriptionFactory` factory the returns `StreamableProvider` for the entity. There are a lot of examples. For example, `DAppLocalSubscriptionFactory`
2) Subscribe to the provider in the TabManager. Now once you update Tab entity the update will automatically arrive to the `TabManager`
3) Make tabs `Observable<[UUID: DAppBrowserTab]> = [:]` and update once you receive changes from the provider
4) Add observers to the state.

That approach avoids a lot of additional logic we have now. Such as fetch tabs and subscribing at the same time (saw in the interactor).

novawallet/Modules/DApp/DAppBrowser/Tabs/WebViewRenderFIlesOperationFactory/WebViewFilesRenderOperationFactory.swift
```diff
@@ -0,0 +1,99 @@
...
+     *      - ids: Idetifier of a webviews for which render data
+     *  must be removed.
+     *
+     *  - Returns: `CompoundOperationWrapper` which produces void
+     *  in case files removed succesfully.
+     */
+    func removeRenders(for ids: [UUID]) -> CompoundOperationWrapper<Void>
+}
+
+extension RuntimeFilesOperationFactory: WebViewRenderFilesOperationFactoryProtocol {
```
ERussel, line 67: `RuntimeFilesOperationFactory` is a separate factory to manage blockchain runtimes. I would create new operation factory for `WebViewRenderFilesOperationFactoryProtocol`.

PR #1320, opened 2024-12-28
Take keep alive cases for xcm transfers

ERussel:
## Purpose

There are chains that requires for account to be alive after xcm transfer. This means the account must leave at least existential deposit. The solution is to introduce `AssetFungibilityPreservationProvider` that defines assets for which xcm transfer logic must ensure to leave ed after transfer: ASTR (Astar chain), all assets on PAH and KAH. The validation itself is integrated to the new cross chain swap logic via operation metadata and to the legacy cross chain transfer logic.


svojsu (approved):
So when a cross-chain swap route goes through PAH/KAH, it will leave ED for the swap's destination asset as a side effect if XCM follows after the PAH/KAH swap, right?

ERussel:
> So when a cross-chain swap route goes through PAH/KAH, it will leave ED for the swap's destination asset as a side effect if XCM follows after the PAH/KAH swap, right?

Currently, PAH/KAH can't be intermediate edge as it requires delivery payment. However I have added filtering of the edges with keep alive requirement in the graph to take into account Astar case.

PR #1331, opened 2025-01-14
Feature/deleted wallet state cleaning

svojsu:
## SUMMARY
Removing a wallet from the app now requires cleaning associated dapp browser state and settings. This PR introduces the `WalletStorageCleaning` interface and implementations to handle this in a cascaded way.

## SOLUTION
The solution is based on a general interface `WalletStorageCleaning` that is used for both the top-level cleaner object and sub-cleaners (we can call them domain cleaners). 

Currently introduced sub-cleaners are `WalletBrowserStateCleaner` and `DAppSettingsCleaner`, with `RemovedWalletStorageCleaner` as the top-level cleaner.

The top-level cleaner is compound and used by `WalletUpdateMediator`.

To extend cleaning behavior, the implementation of the top-level cleaner should not be modified. Instead, we should create a sub-cleaner implementation of the `WalletStorageCleaning` interface and pass it to the top-level cleaner during initialization. We use `WalletStorageCleanerFactory` for this purpose.


novawallet/Common/Storage/WalletStorageCleaning/WalletDeleteStorageCleanerFactory.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+final class WalletStorageCleanerFactory {
+    static func createWalletStorageCleaner(using operationQueue: OperationQueue) -> WalletStorageCleaning {
+        let browserStateCleaner = createBrowserStateCleaner(using: operationQueue)
```
ERussel, line 5: Can we also add cleaner for DAppSettings here?

svojsu: Yep, added

novawalletTests/Common/Storage/WalletUpdateMediatorTests.swift
```diff
@@ -227,14 +238,6 @@ final class WalletUpdateMediatorTests: XCTestCase {
         
         XCTAssertTrue(common.selectedAccountSettings.value.identifier == selectedWallet.identifier)
         
-        // when
-        
-        let wrapper = common.walletUpdateMediator.saveChanges {
-            .init(newOrUpdatedItems: [], removedItems: [removedWallet])
-        }
-        
-        common.operationQueue.addOperations(wrapper.allOperations, waitUntilFinished: true)
```
svojsu, line 236: Not sure why this was implemented this way. We don't need to do this since we call `common.update` immediately afterward.

ERussel: Yep, looks strange. Good spot

novawalletTests/Helper/WalletStorageCleanerFactory.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
+@testable import novawallet
+import Operation_iOS
+
+extension WalletStorageCleanerFactory {
+    static func createTestCleaner(
+        operationQueue: OperationQueue,
+        storageFacade: UserDataStorageTestFacade
```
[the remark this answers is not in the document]
svojsu: We can, but it won't change much since we're using shared DAppBrowserTabManager in the main target, so different factory methods will serve better for such cases

novawallet/Common/Storage/WalletsUpdateMediator.swift
```diff
@@ -227,28 +238,36 @@ extension WalletUpdateMediator: WalletUpdateMediating {
             dependingOn: newSelectedWalletOperation
         )
 
-        selectedWalletUpdateOperation.addDependency(saveOperation)
+        selectedWalletUpdateOperation.addDependency(newSelectedWalletOperation)
```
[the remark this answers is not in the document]
svojsu: Thanks, fixed

ERussel (approved):

PR #1340, opened 2025-01-16
Fix restore collection view layout on rotations

ERussel:
## Purpose

It seems that there is a bug in iOS that prevents UICollectionView to update cell frame on device rotation in some cases like ours with custom presentation. The PR adds a workaround to address the issue by forcing update of the layout when transition animation ends.


ERussel:
Implemented in [PR](https://github.com/novasamatech/nova-wallet-ios/pull/1344)

PR #1349, opened 2025-01-23
Mythos staking setup screen

ERussel:
## Purpose

- Introduce CollatorStaking classes set to reuse logic between Parachain Staking and Mythos Staking:
ViewController for setup/stake more cases, preferred collators logic, view model factory for collators display, common interfaces for selected collators and reward services
- Implement Staking Details sync service to gather full staking information into one model
- Implement Frozen Balance Store to gather all staking locks into one model
- Implement Presenter and Interactor for staking setup/stake more amount screen

Note: validations and staking reward service will be implemented in the separate tasks

##

StakingSetup


novawallet/Modules/Staking/Mythos/MythosStakingSetup/MythosStakingSetupPresenter.swift
```diff
@@ -0,0 +1,522 @@
...
+
+        interactor.applyCollator(with: collatorId)
+    }
+
+    func changeCollator(with collatorId: AccountId, name: String?) {
+        guard let newAddress = try? collatorId.toAddress(using: chainAsset.chain.chainFormat) else {
+            return
+        }
+
+        changeCollator(with: DisplayAddress(address: newAddress, username: name ?? ""))
```
svojsu, line 282: Same over here. I think it's easier to read while we don't initialize struct with multiple params on the same line with function call. WDYT?

ERussel: Yeah, agree. Probably good to think about linting such case, not sure if SwiftLint supports them out of the box

svojsu:
Will check https://github.com/novasamatech/nova-wallet-ios/pull/1353 for fixes

svojsu (approved):

PR #1351, opened 2025-01-24
Feature/popular dapps

svojsu:
### SUMMARY
This PR adds a "Popular DApps" section for the `DAppList` module.

### SOLUTION
- Source popular dApps URLs from the `DAppList` model
- Use the given order and match them with the provided dApp array
- For categories, show randomized popular dApps first, followed by additional randomized dApps
- Implement seeded randomization to ensure consistency for the same input


novawallet/Modules/DApp/DAppList/DAppListPresenter.swift
```diff
@@ -139,6 +144,7 @@ extension DAppListPresenter: DAppListInteractorOutputProtocol {
         }
```
ERussel, line 144: Check above is incorrect. Probably use `if` instead of `guard`

ERussel (approved):

PR #1363, opened 2025-01-31
Fix/webview permissions

svojsu:
- allows inline media playback for webview
- grand media capture permissions
- ask gallery permissions


novawallet/Modules/PayCard/PayCardViewController.swift
```diff
@@ -44,6 +45,7 @@ final class PayCardViewController: UIViewController, ViewHolder {
         super.viewDidLoad()
 
         setupWebView()
+        requestPhotoLibraryAuthorizationIfNeeded()
```
ERussel, line 48: I would try to request access in the web view's delegate method that we added

ERussel (approved):

PR #1364, opened 2025-01-31
Mythos alerts & network info

ERussel:
## Purpose

- add redeem and not active collator alerts
- implement network info data retrieval
- abstract network info view model factory to reuse for Mythos

##
Alerts

NetworkInfo


novawallet/Modules/Staking/CollatorStaking/ViewModel/CollatorStkNetworkInfoViewModelFactory.swift
```diff
@@ -0,0 +1,145 @@
+import Foundation
+import SoraFoundation
+
+struct CollatorStkNetworkModel {
```
svojsu, line 4: I think it's better to move this model to separate file cause it's not a view model. So file won't get messy.

novawallet/Modules/Staking/CollatorStaking/ViewModel/CollatorStkNetworkInfoViewModelFactory.swift
```diff
@@ -0,0 +1,145 @@
...
+    private func createUnstakingPeriodViewModel(
+        unstakingDuration: TimeInterval?
+    ) -> LocalizableResource<String> {
+        if let unstakingDuration {
+            return LocalizableResource { locale in
+                let formattedString = unstakingDuration.localizedDaysHoursOrFallbackMinutes(for: locale)
+                return "~\(formattedString)"
+            }
+        } else {
+            return LocalizableResource { _ in "" }
```
svojsu, line 97: If there is no `unstakingDuration` yet, we will show an empty string instead of `.loading` state. Is it right?

novawallet/Modules/Staking/Operations/MythosStaking/MythosStakingNetworkInfoFactory.swift
```diff
@@ -0,0 +1,126 @@
...
+    let requestFactory: StorageRequestFactory
+    let keysFetchFactory: StorageKeysOperationFactoryProtocol
+
+    init(operationQueue: OperationQueue) {
+        requestFactory = StorageRequestFactory.createDefault(with: operationQueue)
+        keysFetchFactory = StorageKeysOperationFactory(operationQueue: operationQueue)
+    }
+}
+
+extension MythosStkNetworkInfoOperationFactory {
```
svojsu, line 29: not a big deal, but probably should be `private`

novawallet/Modules/Staking/CollatorStaking/ViewModel/CollatorStkNetworkInfoViewModelFactory.swift
ERussel: We don't have loading state here and display empty string in this case

novawallet/Modules/Staking/Operations/MythosStaking/MythosStakingNetworkInfoFactory.swift
ERussel: Will fix the comments in a separate PR just to avoid merge conflicts

PR #1367, opened 2025-02-03
[CU-8697j1xuw] Fix address bar spoofing issue

ERussel:
## Purpose

This PR moves current URL update logic for the DApp browser to the `webView:didCommit:` from the observer to fix url/title spoofing problem. However we still allow to update url from the observer when previous and current origins are the same - some dapps may use js scripts to navigate between pages that prevent delegate call.


stepanLav:
Task linked: [CU-8697j1xuw iOS. Process address bar spoofing](https://app.clickup.com/t/8697j1xuw)

svojsu (approved):

PR #1368, opened 2025-02-03
[CU-8697q62tu] Use fallback price if no pools for fee found

ERussel:
## Purpose

Currently, we have indefinite fee loading state when there is no pool to swap custom token for fee on Hydra. However in this case on chain logic utilises [fallback price](https://github.com/galacticcouncil/hydration-node/blob/v34.0.0/pallets/transaction-multi-payment/src/lib.rs#L98). The PRs adds proxy to the Hydra quote factory to use fallback price in case there are no routes found by actual quote factory

##

feeCalc


stepanLav:
Task linked: [CU-8697q62tu iOS. Process paying custom fees on Hydration for sufficient tokens without liquidity](https://app.clickup.com/t/8697q62tu)

PR #1371, opened 2025-02-05
[CU-8696auqzn] Open polkadot governance by default from the universal link

ERussel:
## Purpose

Currently, universal link requires chainId parameter in it. Current PR allows to open Polkadot by default if there is no chainId in the link.


stepanLav:
Task linked: [CU-8696auqzn iOS. Deep Links. Set Polkadot chain as default for all Deep Links with chain Id](https://app.clickup.com/t/8696auqzn)

svojsu (approved):

PR #1374, opened 2025-02-06
Fix wc signing

ERussel:
## Purpose

Current logic in the `WalletConnectService` disconnects from the socket when viability flag is false but not reconnects to it when flag becomes true. That might be an issue when a user switches between wifi and lte or for some other reasons when flag changes while using the app. The PRs fixes the problem by allowing reconnection only in case socket is currently disconnected.

### DIFF
--- a/novawallet/Common/Services/WalletConnect/WalletConnectService.swift
+++ b/novawallet/Common/Services/WalletConnect/WalletConnectService.swift
@@ -411,11 +411,13 @@ private final class DefaultWebSocket: WebSocketConnecting {
                 )
             case .cancelled:
                 self?.markDisconnectedAndNotify(error: nil)
-            case .reconnectSuggested:
-                self?.protectedRestart()
+            case let .reconnectSuggested(isBetter):
+                if isBetter {
+                    self?.protectedRestartIfDisconnected()
+                }
             case let .viabilityChanged(isViable):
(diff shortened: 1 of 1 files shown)


novawallet/Common/Services/WalletConnect/WalletConnectService.swift
```diff
@@ -437,13 +443,20 @@ private final class DefaultWebSocket: WebSocketConnecting {
...
+        defer {
+            mutex.unlock()
+        }
 
-        mutex.unlock()
+        guard connectionState == .notConnected else {
+            return
+        }
+
+        connectionState = .connecting
```
svojsu, line 457: If I understand correctly, previously we didn't change connection state on `protectedRestart` call. Looks like we could make a private method to hold:

1. Changing connection state to `.connecting`
2. Call `stopWebsocket`
3. Call `startWebsocket`

This way we won't lose any of these next time. WDYT?

ERussel: Don't see any reason to do that since:

- we don't want to call `stopWebsocket` if we are disconnected;
-  better to have logic related to state in one place: otherwise state is checked in one place and modified in another;

svojsu (approved):

PR #1381, opened 2025-02-14
Fix apr calc and more

ERussel:
## Purpose

- APR calculation must account only non invurnerable collators
- Add option to display min number of minutes for time
- Remove restriction for updates on Start Staking based on block numbers

### DIFF
--- a/novawallet/Modules/Staking/Services/RewardCalculatorService/MythosStaking/MythosRewardCalculatorEngine.swift
+++ b/novawallet/Modules/Staking/Services/RewardCalculatorService/MythosStaking/MythosRewardCalculatorEngine.swift
@@ -19,7 +19,6 @@ import BigInt
  */
 final class MythosRewardCalculatorEngine {
     let totalStaked: BigUInt
-    let yearlyEmission: Decimal
     let aprByCollator: [AccountId: Decimal]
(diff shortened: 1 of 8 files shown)


novawallet/Modules/Staking/Services/RewardCalculatorService/MythosStaking/MythosRewardCalculatorEngine.swift
```diff
@@ -31,20 +30,22 @@ final class MythosRewardCalculatorEngine {
 
         let yearlyEmission = Self.calculateYearlyEmission(for: params)
 
+        let rewardableCollatorsCount = params.collators.filter(\.rewardable).count
+
         let aprByCollator: [AccountId: Decimal] = params.collators.reduce(into: [:]) { accum, collator in
-            guard let stake = collator.info?.stake else {
+            guard collator.rewardable, let stake = collator.info?.stake else {
```
svojsu, line 36: As it's already been filtered earlier, I think we can have `rewardableCollators` instead of `rewardableCollatorsCount`, then reduce them with guarding by stake existence, and use `.count` passed to `calculateApr` method. Then it's unnecessary to guard by `rewardable`

svojsu (approved):

PR #1384, opened 2025-02-16
Add support of the block hash in the state call

ERussel:
## Purpose

Currently, state call request is always executed on the last known by the node block. However, there are certain case where one wants to execute query requests on the certain block. For example, such block hash might be received on the state subscription update. The PR introduces an option to provide block hash to the `StateCallRequestFactory`.

### DIFF
--- a/novawallet/Modules/Staking/Services/RewardCalculatorService/RelayChain/PolkadotInflationPredictionFactory.swift
+++ b/novawallet/Modules/Staking/Services/RewardCalculatorService/RelayChain/PolkadotInflationPredictionFactory.swift
@@ -1,0 +1,84 @@
+import Foundation
+import SubstrateSdk
+import Operation_iOS
+
+protocol PolkadotInflationPredictionFactoryProtocol {
+    func createPredictionWrapper(
+        for connection: JSONRPCEngine,
+        runtimeProvider: RuntimeCodingServiceProtocol,
+        at block: BlockHash?
+    ) -> CompoundOperationWrapper<RuntimeApiInflationPrediction>
+}
+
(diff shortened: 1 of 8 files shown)


novawallet/Modules/Staking/Services/RewardCalculatorService/RelayChain/PolkadotInflationPredictionFactory.swift
```diff
@@ -0,0 +1,84 @@
...
+    ) -> CompoundOperationWrapper<RuntimeApiInflationPrediction> {
+        let codingFactoryOperation = runtimeProvider.fetchCoderFactoryOperation()
+
+        let fetchWrapper = OperationCombiningService<RuntimeApiInflationPrediction>.compoundNonOptionalWrapper(
+            operationQueue: operationQueue
+        ) {
+            let codingFactory = try codingFactoryOperation.extractNoCancellableResultData()
+
+            let runtimeApi = try codingFactory.metadata.getRuntimeApiMethodOrError(
+                for: "Inflation",
```
svojsu, line 51: Why don't we add this module and method name to the `StateCallRpc` enum?

svojsu (approved):

novawallet/Modules/Staking/Services/RewardCalculatorService/RelayChain/PolkadotInflationPredictionFactory.swift
ERussel: StateCallRpc not a good place to have each runtime api used in the project. Decided to go with the approach to define the api in the module/service it is used.

PR #1385, opened 2025-02-17
[CU-8697y3wrn] Fix adding evm based network from preconfigured

ERussel:
## Purpose

Currently if a network is added from preconfigured list the options is has only used and derived ones are ignored preventing from cases when evm address support is derived from runtime. The PR introduces the logic that merges the options from preconfigured network and derived network.

Also PR fixes that logic that marks network as evm only it is ethereumBased and has no runtime in the preconfigured network model


stepanLav:
Task linked: [CU-8697y3wrn iOS. Incorrect chain detect from preConfigured](https://app.clickup.com/t/8697y3wrn)

svojsu (approved):

PR #1386, opened 2025-02-17
Feature/calculate height for text

svojsu:
### SUMMARY

This PR introduces the approach to calculate browser module height based on one with larger content.

### DIFF
--- a/novawallet/Modules/Banners/BannersPresenter.swift
+++ b/novawallet/Modules/Banners/BannersPresenter.swift
@@ -83,7 +83,7 @@ extension BannersPresenter: BannersInteractorOutputProtocol {
         localizedResources = updatedLocalizedResources
         provideBanners()
 
-        moduleOutput?.didUpdateContentLocale()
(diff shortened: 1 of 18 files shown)


novawallet/Modules/Banners/BannersPresenter.swift
```diff
@@ -82,6 +82,8 @@ extension BannersPresenter: BannersInteractorOutputProtocol {
     func didReceive(_ updatedLocalizedResources: BannersLocalizedResources?) {
         localizedResources = updatedLocalizedResources
         provideBanners()
+
+        moduleOutput?.didUpdateContentLocale()
```
ERussel, line 86: Not sure why we have Locale in the end. Probably, better to leave just `didUpdateBannersContent`

ERussel (approved):

novawallet/Modules/Banners/BannersPresenter.swift
svojsu: Yes, would be better.

PR #1389, opened 2025-02-17
Feature/banners autoscroll

svojsu:
### SUMMARY

This PR implements auto scrolling logic for new `Banners` module by introducing `AutoScrollManager` object used by view controller.

### SCREEN RECORDS

https://github.com/user-attachments/assets/aa6530f1-9a85-4961-963b-7c40261dd9a0

### DIFF
--- a/novawallet/Modules/Banners/BannersViewController.swift
+++ b/novawallet/Modules/Banners/BannersViewController.swift
@@ -78,8 +78,6 @@ private extension BannersViewController {
 
         rootView.setBackgroundImage(widgetModel.banners.first?.backgroundImage)
         rootView.setCloseButton(available: widgetModel.showsCloseButton)
-
-        autoScrollManager.setupScrolling()
     }
 
     func setupPageControl() {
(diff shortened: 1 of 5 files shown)


novawallet/Modules/Banners/BannersViewController.swift
```diff
@@ -292,13 +321,16 @@ extension BannersViewController: BannersViewProtocol {
             !updatedViewModel.banners.isEmpty
         else { return }
 
+        autoScrollManager.stopScrolling()
```
ERussel, line 324: Shouldn't we stop it before the guard? Otherwise in case of empty banners autoscroll will still be live

novawallet/Modules/Banners/BannersViewController.swift
```diff
@@ -338,6 +370,8 @@ extension BannersViewController: UIScrollViewDelegate {
     func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
         guard let staticState else { return }
 
+        autoScrollManager.stopScrolling()
```
ERussel, line 373: Shouldn't we enable it again in `endDecelerating`?

ERussel (approved):

novawallet/Modules/Banners/BannersViewController.swift
svojsu: As far as I remember, we want to completely stop autoscrolling in case user scrolled it by himself.

novawallet/Modules/Banners/BannersViewController.swift
svojsu: Yes, you're right. Thanks!

PR #1391, opened 2025-02-18
Mythos bug fixes & Improvements

ERussel:
## Purpose

The PR introduces MythosStakingStakeSyncService, a service that maintains an in-memory state to track claimable rewards and staking details. This solution is more robust as it directly subscribes to blockchain storage updates, ensuring synchronization with the latest block. By maintaining state in alignment with the most recent block, the service guarantees that queries for claimable rewards and staking details reflect the correct blockchain state.

In general PR fixes the following issues:

- [CU-8697zgj7w] Remove pending rewards validation from stake more and unstake flow and automatically include ClaimRewards call into batch;
- Collator sync service now using fixed block time that guarantees the same APR as on staking dashboard;
- Fix issue when pending rewards sometimes not updating;
- Transaction errors now are displayed in alert view due to the fact `DispatchCallError` now `ErrorConvertible`;


stepanLav:
Task linked: [CU-8697zgj7w iOS. Add claim rewards to batch during unstake all, stake more](https://app.clickup.com/t/8697zgj7w)

svojsu (approved):

PR #1393, opened 2025-02-19
Add staking rewards support for Mythos

ERussel:
## Purpose

- Refactored transaction history local filters and introduce Mythos filter to get rid of transfers from the system pot account
- Add single staking rewards local data provider to get rid of code duplication for each staking type


novawallet/Modules/TransactionHistory/LocalFilter/Filters/TransactionHistoryTransfersFilter.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+final class TransactionHistoryTransfersFilter {
```
svojsu, line 3: Could be struct so we wouldn't care about finality and init

svojsu (approved):

novawallet/Modules/TransactionHistory/LocalFilter/Filters/TransactionHistoryTransfersFilter.swift
ERussel: I am still old school and prefer classes when there is non trivial logic inside

novawallet/Common/DataProvider/StakingRewardsLocalSubscriptionFactory.swift
```diff
@@ -0,0 +1,63 @@
...
+        endTimestamp: Int64?,
+        api: LocalChainExternalApi,
+        assetPrecision: Int16
+    ) throws -> AnySingleValueProvider<TotalRewardItem> {
+        clearIfNeeded()
+
+        let timeIdentifier = [
+            startTimestamp.map { "\($0)" } ?? "nil",
+            endTimestamp.map { "\($0)" } ?? "nil"
+        ].joined(separator: "-")
```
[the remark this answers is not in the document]
ERussel: Will fix in the next PR

PR #1396, opened 2025-02-20
Feature/basic price chart

svojsu:
### SUMMARY
This PR introduces the complete UI part of the "Price Charts" feature.

### SOLUTION
Charts are built using the `DGCharts` library. The interactor provides stubbed data.
For demonstration, please check the screenshots and recordings.

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/e606bdfa-8eac-4d9a-af41-3cd8b773ef7e
https://github.com/user-attachments/assets/d6d47246-fe6d-4bc0-9d84-38687f18d6db


novawallet/Modules/AssetPriceChart/AssetPriceChartDataSource.swift
```diff
@@ -0,0 +1,180 @@
...
+            let chartModel = widgetViewModel?.chartModel.value,
+            let colors = createColors()
+        else { return nil }
+
+        let currentEntries = chartModel.dataSet
+
+        let entriesBefore = currentEntries.filter { $0.x <= selectedEntry.x }
+        let entriesAfter = Array(currentEntries[entriesBefore.count - 1 ..< currentEntries.count])
+
+        let dataBefore = createChartData(
```
ERussel, line 71: If we even index all the items to optimize the search it will still remain linear as we reload the chart after a finger move as i understand. We might want to look into possible real time optimizations the library provides in terms of redrawing the chart when dataset is not changed - intuitively it should be possible asynchroniously update the color for the already existing pixels in the CPU. However i don't think it is blocking the PR.

novawallet/Modules/AssetPriceChart/Factory/AssetPriceChartViewModelFactory.swift
```diff
@@ -0,0 +1,211 @@
...
+    ) -> PricePeriodChangeViewModel {
+        let periodChangeDecimal = abs(lastPrice - firstPrice)
+
+        let periodChangeAmountText = priceFormatter(priceId: priceData.currencyId)
+            .value(for: locale)
+            .stringFromDecimal(periodChangeDecimal) ?? ""
+
+        let percentText = priceChangePercentFormatter
+            .value(for: locale)
+            .stringFromDecimal(periodChangeDecimal / firstPrice) ?? ""
```
ERussel, line 74: worth to check for zero devision to avoid NaNs

ERussel (approved):

novawallet/Modules/AssetPriceChart/AssetPriceChartDataSource.swift
svojsu: Actually, getting indexes with O(1) still much faster than it's now. Will present in next PR anyway.

novawallet/Modules/AssetPriceChart/Factory/AssetPriceChartViewModelFactory.swift
svojsu: Will fix in the next one

PR #1401, opened 2025-02-21
Fix/optimize data sets splitting

svojsu:
### SUMMARY

This PR adds several fixes and optimizations, identified during the review of previous PRs:
- Adds entry indexing for faster access when splitting chart datasets
- Adds protection against float division by zero by checking `isNaN` and `isInfinite` properties of results
- Fixes implicit self-capture
- Fixes warnings


novawallet/Modules/AssetPriceChart/Factory/AssetPriceChartViewModelFactory.swift
```diff
@@ -76,9 +76,13 @@ private extension AssetPriceChartViewModelFactory {
             .value(for: locale)
             .stringFromDecimal(periodChangeDecimal) ?? ""
 
+        var percent = periodChangeDecimal / firstEntry.value
+
+        percent = (!percent.isNaN && !percent.isInfinite) ? percent : .zero
```
ERussel, line 81: I would just check that `firstEntry.value > 0` before division. Looks safer.

ERussel (approved):

PR #1403, opened 2025-02-21
[CU-86981m4xn] Always generate new wallet with default ethereum derivation path

ERussel:


stepanLav:
Task linked: [CU-86981m4xn iOS. Using not default derivation path for evm accounts when creating wallet in Cloud](https://app.clickup.com/t/86981m4xn)

PR #1404, opened 2025-02-21
update price according to selected entry

svojsu:
### SUMMARY

This PR adds logic for updating current displayed price when user interacts with chart.


ERussel (commented):
As discussed, let use only price chart data for the price chart widget instead of combining with current price

ERussel (approved):

PR #1407, opened 2025-02-25
Independent Browser navigation

svojsu:
### SUMMARY
This PR introduces a new `BrowserNavigation` module that encapsulates all the logic for finding the needed dapp, forming a browser tab, and opening the browser regardless of the current app state.

### SOLUTION
The `BrowserNavigation` module is a simple top-level module that has no view. It stores the current dapps list and contains all the logic needed to open the browser. 
The module is provided by `BrowserNavigationFactory` as a singleton and doesn't need to be injected into any feature modules. Instead, the `BrowserOpening` protocol should be used for feature-module wireframes.


novawallet/Modules/BrowserNavigation/BrowserNavigationFactory.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+import Operation_iOS
+import UIKit
+
+enum BrowserNavigationFactory {
+    static func createNavigation() -> BrowserNavigationProtocol? {
+        guard let mainAppContainer = findMainContainer() else { return nil }
```
ERussel, line 7: I would suggest to check SelectedWalletSettings.shared.value and don't create navigation module if there is no wallet

ERussel (approved):

novawallet/Modules/BrowserNavigation/BrowserNavigationFactory.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    private static func findMainContainer() -> NovaMainAppContainerViewProtocol? {
+        UIApplication
+            .shared
+            .windows
+            .first { $0.isKeyWindow }?
+            .rootViewController as? NovaMainAppContainerViewProtocol
+    }
+
+    static var shared: BrowserNavigationPresenter = {
```
ERussel, line 51: I wouldn't do it singleton. What do you think of attaching it to the lifecycle of the main tab bar?

ERussel (approved):

PR #1409, opened 2025-02-25
Feature/balance widget update

svojsu:
### SUMMARY

This PR introduces the new updated balance widget for `AssetDetails` module.

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/acf3723e-f167-4db5-9028-22ca291de354

### DIFF
--- a/novawallet/Modules/AssetDetails/View/AssetDetailsBalanceWidget.swift
+++ b/novawallet/Modules/AssetDetails/View/AssetDetailsBalanceWidget.swift
@@ -119,12 +119,11 @@ private extension AssetDetailsBalanceWidget {
     }
 
     func setupActions() {
-        let tapGesture = UITapGestureRecognizer(
-            target: self,
-            action: #selector(actionTap)
+        totalCell.addTarget(
+            self,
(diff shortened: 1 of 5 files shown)


novawallet/Modules/AssetDetails/View/AssetDetailsBalanceWidget.swift
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    func setupActions() {
+        let tapGesture = UITapGestureRecognizer(
+            target: self,
+            action: #selector(actionTap)
+        )
+
+        totalCell.addGestureRecognizer(tapGesture)
```
ERussel, line 127: I believe the cell is inherited from RowView and RowView in turn is decedent of the UIControl. Probably, we can add the action directly instead of using gesture recogniser. What do you think?

svojsu: You're absolutely right

ERussel (approved):

PR #1410, opened 2025-02-26
Fix/price provider

svojsu:
### SUMMARY

This PR refactors the price acquisition logic, ditching the need to pass a priceIds set to receive prices. Instead, we now fetch all prices, making the price provider independent of upper-level chain acquisition logic and simplifying the price provider API.
Also, price subscription usage is unified across the app by using `PriceLocalStorageSubscriber`.


novawallet/Modules/AssetList/Base/AssetListBaseInteractor.swift
```diff
@@ -256,58 +256,9 @@ class AssetListBaseInteractor: WalletLocalStorageSubscriber, WalletLocalSubscrip
         baseBuilder?.applyRemovedPriceChainAssets(chainAssetIds)
     }
 
-    private func updatePriceProvider(
-        for priceIdSet: Set<AssetModel.PriceId>,
-        currency: Currency
-    ) {
+    private func updatePriceProvider(currency: Currency) {
         priceSubscription = nil
```
ERussel, line 260: We need `clear(&priceSubscription)` instead of assigning nil

novawallet/Modules/ChainAssetSelection/ChainAssetSelectionInteractor.swift
```diff
@@ -129,59 +129,12 @@ final class ChainAssetSelectionInteractor {
             }
         }
 
-        setupPriceProvider(for: Set(availableTokenPrice.values), currency: selectedCurrency)
+        setupPriceProvider(currency: selectedCurrency)
     }
 
-    private func setupPriceProvider(for priceIdSet: Set<AssetModel.PriceId>, currency: Currency) {
+    private func setupPriceProvider(currency: Currency) {
         priceSubscription = nil
```
ERussel, line 136: We need clear(&priceSubscription) instead of assigning nil

ERussel (commented):
Overall looks good. We need to fix comments

ERussel (approved):

PR #1422, opened 2025-03-04
Feature/price chart entries

svojsu:
### SUMMARY

This PR addresses several issues:

- Implements a 'short' long press gesture to highlight entries, effectively solving the problem of highlighting with disabled taps due to API limitations.
- Adds entry distribution logic to proportionally utilize available width with an optimized dataset. This enables large datasets to display without frame rate drops.

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/7648a5a8-6bd7-49f3-b57e-ca73d8e0ab86


novawallet/Modules/AssetPriceChart/Factory/AssetPriceChartViewModelFactory.swift
```diff
@@ -184,8 +185,19 @@ extension AssetPriceChartViewModelFactory: AssetPriceChartViewModelFactoryProtoc
             selectedPeriod: params.selectedPeriod
         )
 
+        var entries: [PriceHistoryItem] = []
+
+        if let allEntries = params.entries {
+            if allEntries.count > params.availablePoints {
+                entries = allEntries
+                    .distribute(intoChunks: params.availablePoints)
+                    .compactMap(\.first)
```
ERussel, line 194: In that case we might miss the latest price. What do you think is we include first and last prices anyway and select remaining from the distribution?

novawallet/Modules/AssetPriceChart/AssetPriceChartViewController.swift
```diff
@@ -194,6 +235,10 @@ extension AssetPriceChartViewController: AssetPriceChartViewProtocol {
         updatePriceChange()
         updatePrice()
     }
+
+    func chartViewWidth() -> CGFloat {
+        rootView.bounds.width
```
ERussel, line 240: I think we might get into a situation where layout is not completed when requested the width. What do you think about providing just an estimation based on the screen size width and margins when presenter's setup method is called?

novawallet/Common/Extension/Foundation/Array+Distribute.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+extension Array {
+    func distribute(intoChunks chunkCount: Int) -> [[Element]] {
+        guard chunkCount > 0 else { return [] }
+        guard chunkCount < count else { return map { [$0] } }
+
+        let totalElements = count
```
ERussel, line 8: Looks complicated for me. Looks like we are reinventing stride function? Here is an example https://github.com/novasamatech/nova-wallet-ios/blob/develop/novawallet/Common/Ledger/LedgerPathBuilder.swift#L63

novawallet/Modules/AssetPriceChart/AssetPriceChartViewController.swift
svojsu: I think that it might work wrong in case when we embed charts with any other layout rules like no margins or different margins.

ERussel (approved):

PR #1423, opened 2025-03-05
History draggable layout change on balance widget operations

svojsu:
### SUMMARY

This PR adds animated layout changes for the draggable history sheet when balance widget is expanding or collapsing. 
Additionally, to prevent early frame minimization that would be visible to users, we no longer use `compactHeight`. The difference between using `containerSize` vs. `compactHeight` is demonstrated in the screen recordings.

### SCREEN RECORDINGS
Using `compactHeight`
https://github.com/user-attachments/assets/8255f508-c24f-4d62-af31-b5f5e3a94ab7

Using `containerSize`
https://github.com/user-attachments/assets/006f7a7b-7fc0-4afe-a13a-6226fee60d47

### DIFF
--- a/novawallet/Modules/AssetDetails/Container/ContainerViewController.swift
+++ b/novawallet/Modules/AssetDetails/Container/ContainerViewController.swift
@@ -5,7 +5,7 @@ import SoraUI
 class ContainerViewController: UIViewController, AdaptiveDesignable {
     private enum Constants {
         static let minimumBottonInset: CGFloat = 60.0
-        static let contentAnimationDuration: TimeInterval = 0.25
+        static let contentAnimationDuration: TimeInterval = 0.2
         static let draggableChangeDuration: TimeInterval = 0.25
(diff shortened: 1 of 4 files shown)


novawallet/Modules/AssetDetails/Container/ContainerViewController.swift
```diff
@@ -217,21 +225,34 @@ class ContainerViewController: UIViewController, AdaptiveDesignable {
...
 
             let compactOriginY = containerSize.height - preferredContentInsets.bottom
-            let compactHeight = preferredContentInsets.bottom
+
             return CGRect(
                 x: 0.0,
                 y: compactOriginY,
                 width: containerSize.width,
-                height: compactHeight
+                height: containerSize.height
```
ERussel, line 233: Didn't get why we are setting full screen size for the compact mode?

ERussel (approved):

PR #1426, opened 2025-03-06
Fix price chart and asset details bugs

svojsu:
### SUMMARY

This PR contains fixes for bugs found during pre-release testing:

- Use original price decimal to calculate price change. Previously, we used values of `Double` type received from the chart. Due to precision differences, the resulting values were incorrect. To resolve this problem, we now store all prices by timestamp using a dictionary. This way we can retrieve the original price for any selected entry and use it to calculate price change accurately.
Described difference is demonstrated in **SCREEN RECORDINGS** section.
- Add haptic feedback for the long press `began` state.
- Disable highlighting for the balance widget's total cell.
- Hide chevron and disable expanding/collapsing behavior for zero balances.
- Fix header fonts and insets on the `Currency` screen.

Additionally:

- The view's balance update logic has been refactored.
- `AssetDetailsBalanceWidget` now supports common styling API.

### SCREEN RECORDINGS

Before:

https://github.com/user-attachments/assets/81c3da50-22c3-4d0f-955b-934f94fcd05f

Fixed:

https://github.com/user-attachments/assets/e44d8720-0831-4915-81c6-6df5326942e5

### DIFF
--- a/novawallet/Modules/AssetDetails/AssetDetailsPresenter.swift
+++ b/novawallet/Modules/AssetDetails/AssetDetailsPresenter.swift
@@ -40,10 +40,6 @@ final class AssetDetailsPresenter: PurchaseFlowManaging, AssetPriceChartInputOwn
         localizationManager = localizableManager
     }
 
-    private func hasLocks(for balance: AssetBalance, externalBalances: [ExternalAssetBalance]) -> Bool {
-        balance.locked > 0 || !externalBalances.isEmpty
-    }
-
     private func calculateTotalExternalBalances(for externalBalances: [ExternalAssetBalance]) -> BigUInt {
         externalBalances.reduce(0) { $0 + $1.amount }
     }
@@ -58,7 +54,6 @@ final class AssetDetailsPresenter: PurchaseFlowManaging, AssetPriceChartInputOwn
         }
 
         let assetDetailsModel = viewModelFactory.createAssetDetailsModel(
-            balance: balance,
             priceData: priceData,
             chainAsset: chainAsset,
             locale: selectedLocale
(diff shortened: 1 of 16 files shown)


novawallet/Modules/AssetDetails/AssetDetailsPresenter.swift
```diff
@@ -67,33 +67,19 @@ final class AssetDetailsPresenter: PurchaseFlowManaging, AssetPriceChartInputOwn
...
-            value: balance.locked + totalExternalBalances,
-            assetDisplayInfo: chainAsset.assetDisplayInfo,
-            priceData: priceData,
-            locale: selectedLocale
-        )
-
-        view.didReceive(totalBalance: totalBalance)
-        view.didReceive(transferableBalance: transferableBalance)
-
-        let isSelectable = hasLocks(for: balance, externalBalances: externalAssetBalances)
```
ERussel, line 94: Do we need the `hasLocks` function now?

novawallet/Modules/AssetDetails/View/AssetDetailsViewModelFactory.swift
```diff
@@ -113,6 +102,59 @@ final class AssetDetailsViewModelFactory: AssetDetailsViewModelFactoryProtocol {
...
+
+        return AssetDetailsBalanceModel(
+            total: totalModel,
+            locked: lockedModel,
+            transferrable: models[2]
+        )
+    }
+
+    func createAssetDetailsModel(
+        balance _: AssetBalance,
```
ERussel, line 139: We don't need balance here as I see

PR #1427, opened 2025-03-07
Hide chart if price is not available

svojsu:
### SUMMARY

This PR adds state-notifying logic for module output. Using this state, the parent module determines if the chart should be shown or not.

### DIFF
--- a/novawallet/Modules/AssetPriceChart/AssetPriceChartPresenter.swift
+++ b/novawallet/Modules/AssetPriceChart/AssetPriceChartPresenter.swift
@@ -57,14 +57,24 @@ private extension AssetPriceChartPresenter {
 
         view?.update(with: viewModel)
     }
+
+    func notifyIfChartAvailable() {
+        guard
+            let priceData,
+            let prices,
+            !prices.isEmpty
+        else { return }
+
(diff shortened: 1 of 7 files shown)


novawallet/Modules/AssetPriceChart/AssetPriceChartPresenter.swift
```diff
@@ -108,22 +118,39 @@ extension AssetPriceChartPresenter: AssetPriceChartPresenterProtocol {
...
             values.forEach { acc[$0.startedAt] = $0.value }
         }
 
+        notifyIfChartAvailable()
         provideViewModel()
     }
 
     func didReceive(price: PriceData?) {
+        guard let price else {
+            moduleOutput?.didReceiveChartState(.unavailable)
```
ERussel, line 137: The problem here is that we can have `nil` price when it is loading. I would introduce another method for communicating price availability by checking whether an asset has price id

svojsu: Good point. Actually, I don't think we need separate method, because interactor is able to communicate `.missingPriceId` error to presenter.

PR #1446, opened 2025-03-14
Feature/history for legacy address

svojsu:
### SUMMARY

This PR adds support for legacy address prefix to address conversion logic.

It also refactors the history fetching logic by modifying `WalletRemoteHistoryFactoryProtocol` and its implementations to use `accountId` and `chainFormat` instead of directly using addresses. The main goal is to ensure substrate factories use `accountId` and `chainFormat` to create legacyAddress without exposing this implementation detail to ethereum factories.

### DIFF
--- a/novawallet/Common/Network/Etherscan/ERC20/EtherscanERC20OperationFactory.swift
+++ b/novawallet/Common/Network/Etherscan/ERC20/EtherscanERC20OperationFactory.swift
@@ -5,13 +5,26 @@ import SoraFoundation
 final class EtherscanERC20OperationFactory: EtherscanBaseOperationFactory {
     let contractAddress: AccountAddress
 
-    init(contractAddress: AccountAddress, baseUrl: URL, chainId: ChainModel.Id) {
+    init(
+        contractAddress: AccountAddress,
+        baseUrl: URL,
+        chainId: ChainModel.Id,
+        operationManager: OperationManagerProtocol
+    ) {
         self.contractAddress = contractAddress
 
-        super.init(baseUrl: baseUrl, chainId: chainId)
+        super.init(
+            baseUrl: baseUrl,
+            chainId: chainId,
+            operationManager: operationManager
+        )
     }
+}
 
-    private func buildUrl(for info: EtherscanERC20HistoryInfo) -> URL? {
+// MARK: Private
+
+private extension EtherscanERC20OperationFactory {
+    func buildUrl(for info: EtherscanERC20HistoryInfo) -> URL? {
         guard var components = URLComponents(url: baseUrl, resolvingAgainstBaseURL: false) else {
             return nil
         }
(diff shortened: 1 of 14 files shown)


novawallet/Common/Network/Subquery/SubqueryHistoryOperationFactory.swift
```diff
@@ -137,81 +148,148 @@ final class SubqueryHistoryOperationFactory {
...
-                     \(poolRewardField)
-                     \(swapField)
+    ) -> BaseOperation<String> {
+        ClosureOperation { [weak self] in
+            guard let self else { throw BaseOperationError.parentOperationCancelled }
+
+            var address = try accountId.toAddress(using: chainFormat)
+            var legacyAddress = try address.toLegacySubstrateAddress(for: chainFormat)
+
+            if ethereumBased {
```
ERussel, line 163: I think we don't need ethereumBased since chain format has enough information. What do you think?

novawallet/Common/Network/Etherscan/ERC20/EtherscanERC20OperationFactory.swift
```diff
@@ -30,44 +43,91 @@ final class EtherscanERC20OperationFactory: EtherscanBaseOperationFactory {
...
+
     func createFetchWrapper(
-        for info: EtherscanERC20HistoryInfo
+        for accountId: AccountId,
+        chainFormat: ChainFormat,
+        pagination: EtherscanPagination
     ) -> CompoundOperationWrapper<WalletRemoteHistoryData> {
-        guard let url = buildUrl(for: info) else {
-            return CompoundOperationWrapper.createWithError(NetworkBaseError.invalidUrl)
+        let infoOperation = createInfoOperation(
```
ERussel, line 75: Don't see any purpose to wrap that into an operation and add dependency on operation manager

novawallet/Common/Network/Etherscan/Native/EtherscanNativeOperationFactory.swift
```diff
@@ -73,33 +128,34 @@ final class EtherscanNativeOperationFactory: EtherscanBaseOperationFactory {
...
-        let context = EtherscanHistoryContext(context: pagination.context ?? [:], defaultOffset: pagination.count)
-        return context.isComplete
-    }
+// MARK: WalletRemoteHistoryFactoryProtocol
 
+extension EtherscanNativeOperationFactory: WalletRemoteHistoryFactoryProtocol {
     func createOperationWrapper(
-        for address: String,
+        for accountId: AccountId,
+        chainFormat: ChainFormat,
```
ERussel, line 136: What do you think if we provide chain into init instead of making dependency on chain format?

svojsu: I don't get the point why chain is better here. I don't think we need anything else from chain model, but chainFormat.

novawallet/Common/Network/Subquery/SubqueryHistoryOperationFactory.swift
svojsu: good point, chainFormat is enough here.

ERussel (approved):

PR #1459, opened 2025-03-18
Refactor xcm fee logic

ERussel:
## Purpose

- extend XCM sync service to sync both legacy and dynamic transfers
- move xcm call creation to `XcmCallDerivation`
- introduce `XcmTransfers` to incapsulate both legacy and dynamic transfers
- introduce `XcmTransferMetadata` to pass additional information about transfer to the XcmTransferService
- move legacy fee logic into `XcmLegacyCrosschainFeeCalculator`
- introduce `DryRunOperationFactory` for call execution
- introduce `CallBuilder` to simplify call construction


novawalletIntegrationTests/XcmTransfers/DryRunBasedFeeTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        )
+        
+        operationQueue.addOperations(callWrapper.allOperations, waitUntilFinished: true)
+        
+        let callCollecting = try callWrapper.targetOperation.extractNoCancellableResultData()
+        
+        var callBuilder: RuntimeCallBuilding = RuntimeCallBuilder()
+        
+        callBuilder = try callCollecting
+            .addingToCall(builder: callBuilder)
```
svojsu, line 78: Why don't we drop `RuntimeCallCollecting` protocol and don't use straight RuntimeCallCollector struct? Are we expecting more `RuntimeCallCollecting` conformances in the future?
It looks like it could be easier to pass `callCollector.call` to `callBuilder.addingLast(call)` instead of messing with re-setting `callBuilder` via `callCollecting. addingToCall` method.
Same applies for `addingToExtrinsic`.

svojsu (approved):

PR #1462, opened 2025-03-18
Feature asset id dapp

ERussel:
## Purpose 

- allow fee in custom asset using `assetId` field for DApp browser

Simulator Screenshot - iPhone 16 Pro Max - 2025-03-18 at 17 39 38

### DIFF
--- a/novawallet/Modules/DApp/DAppOperationConfirm/DAppOperationConfirmInteractor+Proccessing.swift
+++ b/novawallet/Modules/DApp/DAppOperationConfirm/DAppOperationConfirmInteractor+Proccessing.swift
@@ -56,19 +56,12 @@ extension DAppOperationConfirmInteractor {
             }
 
             guard chain.hasAssetHubFees else {
-                throw DAppOperationConfirmInteractorError.extrinsicBadField(
-                    name: "Unsupported fee asset in \(chain.name)"
-                )
+                return nil
             }
 
-            let remoteAsset = try StatemineAssetSerializer.decodeFeeAssetId(
-                assetId,
-                codingFactory: codingFactory
-            )
-
             guard
-                let localAsset = AssetHubTokensConverter.convertToLocalAsset(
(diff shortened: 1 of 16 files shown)


novawallet/Modules/DApp/DAppOperationConfirm/DAppOperationConfirmInteractor+Proccessing.swift
```diff
@@ -42,6 +42,38 @@ extension DAppOperationConfirmInteractor {
...
+            }
+
+            guard
+                let localAsset = AssetHubTokensConverter.convertRawFeeAssetToLocalAsset(
+                    assetId,
+                    on: chain,
+                    using: codingFactory
+                ) else {
+                throw DAppOperationConfirmInteractorError.extrinsicBadField(
+                    name: "Fee asset decoding failed"
```
svojsu, line 69: The error might be missleading. DAppOperationConfirmInteractorError.extrinsicBadField is used to pass bad field name. It might be better to introduce separate error case for this one.

novawallet/Modules/DApp/DAppOperationConfirm/DAppOperationConfirmInteractor+Proccessing.swift
```diff
@@ -42,6 +42,38 @@ extension DAppOperationConfirmInteractor {
...
+    ) -> BaseOperation<ChainAsset?> {
+        ClosureOperation {
+            let extrinsic = try extrinsicOperation.extractNoCancellableResultData()
+            let codingFactory = try codingFactoryOperation.extractNoCancellableResultData()
+
+            guard let assetId = extrinsic.assetId else {
+                return nil
+            }
+
+            guard chain.hasAssetHubFees else {
```
svojsu, line 58: Probably should be guarded combined with `extrinsic.assetId`

svojsu (approved):

PR #1473, opened 2025-03-24
v9.6.0 Unified Address Format support

svojsu:


svojsu:
Release severity: Major
Release version: 9.6.0
Release notes:

Unified Address Formats are here in Nova Wallet v9.6.0! 

- Unified Address Formats (UAF) are now supported for networks that transitioned to UAF
- Added bottom sheet with context of UAF when user tries to copy the address of the networks that transitioned to UAF
- Added a button on Receive screen to view both UAF and Legacy addresses for the networks that transitioned to UAF
- Added support in action flows (Send, etc) for both UAF and Legacy addresses for the networks that transitioned to UAF

ERussel (approved):

PR #1483, opened 2025-03-31
Improve dynamic xcms

ERussel:
## Purpose

The PR fixes issue from task [CU-869898auu]:

- add support for xcmV5 to properly handle dry run on Mythos chain
- determine Multilocation for the sending asset reserve based on `AbsoluteLocation` for reserve and origin chains. Previously it was determined based on reserve and origin chain ids that didn't take into account corner cases.
- handle special case when native asset is sent between system chains (DOT Polkadot Collectives -> Polkadot Asset Hub). In this case this must be remote reserve transfer instead of non reserve transfer.
- get rid of detecting result xcm after dry run based on SentXcm event that non needed now since the issue with forwardedXcms is fixed on chain


stepanLav:
Task linked: [CU-869898auu iOS. XCM. XCM routes auto-calculation](https://app.clickup.com/t/869898auu)

svojsu (approved):

PR #1489, opened 2025-04-02
Introduce execute call builder

ERussel:
## Purpose

The PR introduces `XcmExecuteDerivator` a new call builder for xcm transactions that builds xcm program manually for each type of the transfer. This allow to pay all the fees from the amount being sent. Previously we had an issue that one can't pay delivery fee from the sending amount and also can't send all amount from the account.

Note: Will refactor xcm program construction in a separate PR by moving the logic from the call builder to the new xcm builder entity.

### DIFF
--- a/novawallet/Common/Services/ExtrinsicService/Substrate/Xcm/XcmExecuteDerivator.swift
+++ b/novawallet/Common/Services/ExtrinsicService/Substrate/Xcm/XcmExecuteDerivator.swift
@@ -1,5 +1,10 @@
 import Foundation
 import Operation_iOS
+import SubstrateSdk
+
+enum XcmExecuteDerivatorError: Error {
+    case messageWeightFailed(JSON)
+}
 
 final class XcmExecuteDerivator {
     enum TransferType {
@@ -10,9 +15,17 @@ final class XcmExecuteDerivator {
     }
 
     let chainRegistry: ChainRegistryProtocol
+    let xcmPaymentFactory: XcmPaymentOperationFactoryProtocol
+    let metadataFactory: XcmPalletMetadataQueryFactoryProtocol
 
-    init(chainRegistry: ChainRegistryProtocol) {
+    init(
+        chainRegistry: ChainRegistryProtocol,
+        xcmPaymentFactory: XcmPaymentOperationFactoryProtocol,
+        metadataFactory: XcmPalletMetadataQueryFactoryProtocol
+    ) {
         self.chainRegistry = chainRegistry
+        self.xcmPaymentFactory = xcmPaymentFactory
+        self.metadataFactory = metadataFactory
     }
 }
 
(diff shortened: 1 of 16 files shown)


novawallet/Common/Services/ExtrinsicService/Substrate/Xcm/XcmExecuteDerivator.swift
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+
+extension XcmExecuteDerivator {
+    func half(asset: XcmV4.Multiasset) -> XcmV4.Multiasset {
+        switch asset.fun {
+        case let .fungible(amount):
+            return XcmV4.Multiasset(assetId: asset.assetId, amount: amount / 2)
+        case .nonFungible:
+            return asset
```
svojsu, line 38: no need for return

novawallet/Common/Services/ExtrinsicService/Substrate/Xcm/XcmExecuteDerivator.swift
```diff
@@ -0,0 +1,362 @@
...
+        request.origin.parachainId.isRelayOrSystemParachain &&
+            request.destination.parachainId.isRelayOrSystemParachain &&
+            request.origin.chainAsset.isUtilityAsset
+    }
+
+    func determineTransferType(
+        for request: XcmUnweightedTransferRequest
+    ) -> TransferType {
+        if isTeleport(request: request) {
+            return .teleport
```
svojsu, line 52: no need for return

novawallet/Common/Substrate/Operations/DryRun/DryRunOperationFactory.swift
```diff
@@ -17,45 +17,7 @@ protocol DryRunOperationFactoryProtocol {
...
-                runtimeProvider: runtimeProvider,
-                connection: connection,
-                operationQueue: operationQueue
-            )
-        } catch {
-            return CompoundOperationWrapper.createWithError(error)
-        }
-    }
-}
+final class DryRunOperationFactory: SubstrateRuntimeApiOperationFactory {}
```
svojsu, line 20: What do you think about going for composition over inheritance here so we don't sum the complexity of base and child classes?

novawallet/Common/Substrate/Operations/XcmPaymentApi/XcmPaymentOperationFactory.swift
```diff
@@ -0,0 +1,37 @@
...
+import Operation_iOS
+
+protocol XcmPaymentOperationFactoryProtocol {
+    func queryMessageWeight(
+        for message: Xcm.Message,
+        chainId: ChainModel.Id
+    ) -> CompoundOperationWrapper<XcmPayment.WeightResult>
+}
+
+class XcmPaymentOperationFactory: SubstrateRuntimeApiOperationFactory {}
```
svojsu, line 12: What do you think about going for composition over inheritance here so we don't sum the complexity of base and child classes?

novawallet/Common/Substrate/Operations/DryRun/DryRunOperationFactory.swift
ERussel: I would keep it as it is because mainly the purpose is that the child class will just provide the parameters and call path and the base class already contains sending logic. Initialisation is also inherited and if we turn to the composition then we should think about initialisation. Finally, I don't see that base class will be changed any time soon instead we just create new runtime apis as child classes and implement proper routing parameters logic

novawallet/Common/Substrate/Operations/XcmPaymentApi/XcmPaymentOperationFactory.swift
ERussel: Same as above

svojsu (approved):

PR #1495, opened 2025-04-08
Feature/on ramp fixes and navigation

svojsu:
### SUMMARY

The PR adds following things:

- Adds same hook-based logic for on-ramp flows.
- Changes navigation so `AssetList` and `AssetDetails` modules are now responsible for managing ramp flow and navigating after ramp completed.
- Adds more logic for handling Transak events.
- Adds alert for unsupported Buy/Sell case.


novawallet/Common/Protocols/Ramp/RampPresentable.swift
```diff
@@ -37,8 +37,18 @@ extension RampPresentable {
         ) else {
             return
         }
-        rampView.controller.modalPresentationStyle = .fullScreen
-        view?.controller.present(rampView.controller, animated: true)
+
+        if view?.controller.isBeingPresented == true {
```
ERussel, line 41: Didn't get why presentation method depends on whether current controller presentation is in progress or completed. It might be a typo so and we just wanted to check whether current controller is modally presented.

novawallet/Common/RampProvider/RampAggregator.swift
```diff
@@ -37,11 +37,11 @@ extension RampAggregator: RampProviderProtocol {
         providers.flatMap { $0.buildRampActions(for: chainAsset, accountId: accountId) }
     }
 
-    func buildOffRampHooks(
+    func buildRampHooks(
         for action: RampAction,
         using params: OffRampHookParams,
-        for delegate: OffRampHookDelegate
+        for delegate: OffRampHookDelegate & OnRampHookDelegate
```
ERussel, line 43: I would typealis `RampHookDelegate = OffRampHookDelegate & OnRampHookDelegate` for simplicity. WDYF?

novawallet/Common/RampProvider/RampProviderProtocol.swift
```diff
@@ -87,20 +88,31 @@ protocol OffRampHookFactoryProvider {
     var offRampHookFactory: OffRampHookFactoryProtocol { get }
 }
 
-extension RampProviderProtocol where Self: BaseURLStringProvider, Self: OffRampHookFactoryProvider {
-    func buildOffRampHooks(
+protocol OnRampHookFactoryProvider {
```
ERussel, line 91: I think `OnRampHookFactoryProviding` or `OnRampHookFactoryProviderProtocol` fits better as we usually use nouns like `OnRampHookFactoryProvider` for concrete implementations of a protocol

novawallet/Modules/AssetDetails/AssetDetailsPresenter.swift
```diff
@@ -152,6 +150,15 @@ extension AssetDetailsPresenter: AssetDetailsPresenterProtocol {
     }
 
     func handleBuySell() {
+        guard availableOperations.rampAvailable() else {
```
ERussel, line 153: I thought we disable buy/sell button if there is no any ramp action. Did this change?

novawallet/Modules/AssetDetails/AssetDetailsWireframe.swift
```diff
@@ -117,4 +117,33 @@ extension AssetDetailsWireframe: AssetDetailsWireframeProtocol {
         let alertController = ModalAlertFactory.createMultilineSuccessAlert(message)
         view?.controller.present(alertController, animated: true)
     }
+
+    func showRampNotSupported(
```
ERussel, line 121: Don't we want to move it to the RampPresentable? I imagine that we want to show the message on the Asset List also.

ERussel (commented):
I am right that now we don't rely on the deeplink but rather derive on ramp completion operation directly from the web page events?

novawallet/Modules/Ramp/Hook/Transak/OffRamp/TransakOffRampResponseHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+                let model = PayCardTopupModel(
+                    chainAsset: chainAsset,
+                    amount: transferData.cryptoAmount,
+                    recipientAddress: transferData.cryptoPaymentData.paymentAddress
+                )
+
+                delegate?.didRequestTransfer(from: model)
+            }
+        } else if
+            let transferEvent = try? JSONDecoder().decode(
```
ERussel, line 51: What to you think about moving decoding logic to the TransakEvent itselt. As i understand event id field is always there and data field depends on eventId. So we could make TransakEvent as enum and implement custom decoding that decodes eventId first and then decodes data based on event id. Finally, once we associate decoded data with the enum case

novawallet/Modules/AssetDetails/AssetDetailsWireframe.swift
svojsu: AssetList picks chainAsset from the list that been filtered at the moment so the only case we can see the message on AssetList is there is no assets with ramp providers. Do you think we should should cover the case?

novawallet/Modules/AssetDetails/AssetDetailsPresenter.swift
svojsu: Yeah, we do not disable it anymore, but styling as disabled. We need the button to be active to show the alert if user taps it.

novawallet/Common/RampProvider/RampAggregator.swift
svojsu: Yeah, I'm agree

svojsu:
> I am right that now we don't rely on the deeplink but rather derive on ramp completion operation directly from the web page events?

Right

ERussel (approved):

PR #1505, opened 2025-04-11
[CU-86983uh2z] Fallback to default custom fee calculation when swaps not available

ERussel:
## Purpose

Currently, if swaps on the PAH are not available the attempt to calculate fee in the custom asset for cross chain transfer on PAH during the swap leads to the error. The PR introduces AssetExchangeFeeEstimatingRouter that fallbacks to the default fee calculation method if swaps are not available in the graph.


stepanLav:
Task linked: [CU-86983uh2z iOS. Not calculating fee for swaps with USDC when user has no native token to pay fee](https://app.clickup.com/t/86983uh2z)

svojsu (approved):

PR #1509, opened 2025-04-15
save last evm network

svojsu:
### SUMMARY
The PR adds logic to persist default evm chain to settings upon addEthereumChain switchEthereumChain metamask requests.


novawallet/Modules/DApp/DAppBrowser/StateMachine/MetamaskStates/DAppMetamaskBaseState.swift
```diff
@@ -75,11 +86,13 @@ class DAppMetamaskBaseState {
         }
 
         let ethereumChain = extendingMetamaskChainWithSubstrate(
-            MetamaskChain.etheremChain,
+            MetamaskChain.ethereumChain,
```
ERussel, line 89: I think we now can safely get the chain by the chain id from the data source instead of allowing to switch only to the Ethereum chain. WDYT?

PR #1510, opened 2025-04-15
Fix long WC request time after WiFi/LTE change

svojsu:
### SUMMARY
The PR fixes long WC request time after WiFi/LTE change by calling forced connection restart if reconnection suggested by WS engine.

### DIFF
--- a/novawallet/Common/Services/WalletConnect/WalletConnectService.swift
+++ b/novawallet/Common/Services/WalletConnect/WalletConnectService.swift
@@ -407,7 +407,7 @@ private final class DefaultWebSocket: WebSocketConnecting {
             markDisconnectedAndNotify(error: nil)
         case let .reconnectSuggested(isBetter):
             if isBetter {
-                restart()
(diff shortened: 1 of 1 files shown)


novawallet/Common/Services/WalletConnect/WalletConnectService.swift
```diff
@@ -459,6 +459,18 @@ private final class DefaultWebSocket: WebSocketConnecting {
...
+    private func restart() {
+        mutex.lock()
+
+        defer {
+            mutex.unlock()
+        }
+
+        connectionState = .connecting
+
+        startWebsocket()
```
ERussel, line 471: I would call the method `protectedForceRestart()` and also call `stopWebsocket()` before starting new one

ERussel (commented):
I would also check the case when there is a pending request and the client reconnects

ERussel (approved):
Let's merge it to a separate branch first and provide a build for extensive testing

PR #1513, opened 2025-04-17
Hide network cell for bytes sign confirmation

svojsu:


novawallet/Modules/DApp/DAppOperationConfirm/ViewModel/DAppOperationBytesConfirmViewModelFactory.swift
```diff
@@ -0,0 +1,5 @@
+final class DAppOperationBytesConfirmViewModelFactory: DAppOperationConfirmViewModelFactory {
```
ERussel, line 1: I would make Base class and 2 implementations that override base class just two avoid cases where we are changing something for extrinsics forgetting that bytes inherit it.

ERussel (approved):

PR #1515, opened 2025-04-25
Add support delegate_with_autocompound for Parachain Staking [CU-8698ghttz]

ERussel:
## Purpose

The deprecated `delegate` call has been removed from [Parachain Staking pallet](https://github.com/moonbeam-foundation/moonbeam/pull/3215). The PR introduces support for `delegate_with_autocompound_call`. Since UI doesn't support autocompound yet the values are set to zero correspondingly. Also PR ensures backward compatability for chains which haven't upgrade yet.

##

Simulator Screenshot - iPhone 16 Pro - 2025-04-25 at 11 54 31

Simulator Screenshot - iPhone 16 Pro - 2025-04-25 at 11 54 53

Simulator Screenshot - iPhone 16 Pro - 2025-04-25 at 11 55 18

### DIFF
--- a/novawallet/Modules/Staking/Parachain/ParaStkStakeConfirm/ParaStkStakeConfirmInteractor.swift
+++ b/novawallet/Modules/Staking/Parachain/ParaStkStakeConfirm/ParaStkStakeConfirmInteractor.swift
@@ -183,38 +183,13 @@ final class ParaStkStakeConfirmInteractor: RuntimeConstantFetching {
             self.extrinsicSubscriptionId = nil
         }
     }
-}
 
-extension ParaStkStakeConfirmInteractor: ParaStkStakeConfirmInteractorInputProtocol {
-    func setup() {
-        feeProxy.delegate = self
-
-        subscribeAccountBalance()
-        subscribePriceIfNeeded()
-        subscribeDelegator()
-        subscribeCollatorMetadata()
-        subscribeScheduledRequests()
-
-        provideMinTechStake()
-        provideMinDelegationAmount()
-        provideMaxDelegationsPerDelegator()
-        provideStakingDuration()
-    }
-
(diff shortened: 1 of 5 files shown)


stepanLav:
Task linked: [CU-8698ghttz Westend direct staking. Inactive stake status with active validators without stake assignment](https://app.clickup.com/t/8698ghttz)

novawallet/Modules/Staking/Parachain/ParaStkStakeConfirm/ParaStkStakeConfirmInteractor.swift
```diff
@@ -249,6 +224,55 @@ extension ParaStkStakeConfirmInteractor: ParaStkStakeConfirmInteractorInputProto
...
+        provideMinDelegationAmount()
+        provideMaxDelegationsPerDelegator()
+        provideStakingDuration()
+    }
+
+    func estimateFee(with callWrapper: DelegationCallWrapper) {
+        let identifier = callWrapper.extrinsicId()
+
+        runtimeProvider.fetchCoderFactory(
+            runningIn: OperationManager(operationQueue: operationQueue),
```
svojsu, line 247: It would be better to pass the `OperationManager`. The only current use of `operationQueue` directly is to enqueue operations in `provideStakingDuration`, and this can also be accomplished using the `OperationManager`.

svojsu (approved):

novawallet/Modules/Staking/Parachain/ParaStkStakeConfirm/ParaStkStakeConfirmInteractor.swift
ERussel: We are planing to deprecate OperationManager and there is no overhead here. So i would leave as it is

PR #1521, opened 2025-05-07
Feature/discover multisig wallets

svojsu:
### SUMMARY
This PR refactors `ProxySyncService` and `ProxySyncChainService`, extending their responsibilities to also sync Multisig accounts that have currently added meta accounts as signatories.
It also adds a `Multisig` entity for persistence and model. This entity relates to either `MetaAccount` or `ChainAccount`, depending on the type of multisig.


novawallet/Common/Services/DelegatedAccounts/Repositories/MultisigAccountsRepository.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        let fetchFactory = SubqueryMultisigsOperationFactory(
+            url: apiURL
+        )
+
+        let fetchOperation = fetchFactory.createDiscoverMultisigsOperation(for: nonCachedSignatories)
+
+        let mapOperation = ClosureOperation<[AccountId: [DelegatedAccount]]> { [weak self] in
+            guard let self else {
```
ERussel, line 51: I don't like such implicit assumptions that operation factory should stay alive till operation completes. Better to catch strongly i think

novawallet/Common/Services/DelegatedAccounts/Repositories/ProxyAccountsRepository.swift
```diff
@@ -0,0 +1,82 @@
...
+                connection: connection,
+                runtimeProvider: runtimeProvider,
+                at: blockHash
+            )
+        } else {
+            .createWithResult(proxies)
+        }
+
+        let mapOperation = ClosureOperation { [weak self] in
+            guard let self else { throw BaseOperationError.parentOperationCancelled }
```
ERussel, line 63: Same is here, catch strongly

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChainSyncService.swift
```diff
@@ -0,0 +1,407 @@
...
+            let chainMetaAccounts = try metaAccountsWrapper.targetOperation.extractNoCancellableResultData()
+
+            let possibleProxiedList: [AccountId] = chainMetaAccounts.compactMap { wallet in
+                guard wallet.info.type != .proxied else {
+                    return nil
+                }
+
+                return wallet.info.fetch(for: chainModel.accountRequest())?.accountId
+            }
+            let possibleMultisigList: [AccountId] = chainMetaAccounts.compactMap { wallet in
```
ERussel, line 241: I think you don't need delegated account wallets in the initial set at all since you want to detect also the case when the delegate is cancelled

ERussel (commented):
I would spent time to work on the following:

- We have initially discussed that the `DelegatedAccountChainSyncService` must depend on set of datasource to fetch delegated accounts and be agnostic on which type of the account it is dealing with. The same i think goes to the changes calculator. Currently i see that both classes manually handle in place proxy and multisigs. That creates a lot of mess. 
The both classes should only know about DelegatedAccountProtocol that exposes account id and possibly identifier. Also we need a separate factory that creates MetaAccountModel from delegated account. Ideally it should be it. Maybe there will be additional classes to differentiate between manually between proxies and multisigs but not the whole syncing logic because it will be hard to maintain it and add additional types of delegated accounts in future.

- Regarding the uniqueness of MetaAccountModel for the universal multisigs. I imagine a class DelegatedAccountSyncBarrier that is passed to each chain sync service it maintains the set of delegated account wallets detected in the current session across all the instances. Each chain sync service then filters the changes is has using barier service and passes to update mediator the result.

novawallet/Common/Services/CloudBackup/Model/CloudBackupFileModelConverter.swift
```diff
@@ -97,7 +97,7 @@ final class CloudBackupFileModelConverter {
             return .polkadotVault
         case .genericLedger:
             return .genericLedger
-        case .proxied:
+        case .proxied, .multisig: // TODO: Verify if backups is available for multisig
```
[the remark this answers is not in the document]
svojsu: Yes, I thought the same. TODO is resolved in the next PR.

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedMetaAccountFactory/MultisigMetaAccountFactory.swift
```diff
@@ -0,0 +1,236 @@
...
+            otherSignatories: discoveredMultisig.otherSignatories(than: signatory),
+            threshold: discoveredMultisig.threshold,
+            status: .new
+        )
+
+        if signatoryWallet.info.chainAccounts.isEmpty {
+            if let substrateAccountId = signatoryWallet.info.substrateAccountId {
+                return .universalSubstrate(substrateAccountId, multisigModel)
+            } else if let ethereumAddress = signatoryWallet.info.ethereumAddress {
+                return .universamEvm(ethereumAddress, multisigModel)
```
ERussel, line 73: .universalEvm

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedMetaAccountFactory/MultisigMetaAccountFactory.swift
```diff
@@ -0,0 +1,236 @@
...
+            let multisig = delegatedAccount as? DiscoveredMultisig,
+            let multisigAccountType = metaAccount.info.multisigAccount()
+        else { return false }
+
+        return switch multisigAccountType {
+        case let .singleChain(chainAccount, multisigModel):
+            chainAccount.accountId == delegatorAccountId &&
+                multisigModel.accountId == multisig.accountId
+        case let .universal(multisigModel):
+            metaAccount.info.substrateAccountId == delegatorAccountId &&
```
ERussel, line 196: what if it is unversal evm?

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedMetaAccountFactory/MultisigMetaAccountFactory.swift
```diff
@@ -0,0 +1,236 @@
...
+                chainAccounts: [],
+                type: .multisig,
+                multisig: multisigModel
+            ))
+        case let .universamEvm(address, multisigModel):
+            ManagedMetaAccountModel(info: MetaAccountModel(
+                metaId: UUID().uuidString,
+                name: name,
+                substrateAccountId: nil,
+                substrateCryptoType: cryptoType.rawValue,
```
ERussel, line 132: I think it should be nil

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountSourcesFactory.swift
```diff
@@ -0,0 +1,49 @@
...
+            let proxyRepository = ChainProxyAccountsRepository(
+                requestFactory: requestFactory,
+                connection: connection,
+                runtimeProvider: runtimeProvider,
+                blockHash: blockHash
+            )
+            sources.append(proxyRepository)
+        }
+
+        let multisigRepository = MultisigAccountsRepository(chain: chain)
```
ERussel, line 44: Shouldn't we also have onchain multisig source to detect based on the events?

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountSyncBarrier.swift
```diff
@@ -0,0 +1,46 @@
...
+extension DelegatedAccountSyncBarrier: DelegatedAccountSyncBarrierProtocol {
+    func filter(_ changes: SyncChanges<ManagedMetaAccountModel>) -> SyncChanges<ManagedMetaAccountModel> {
+        mutex.lock()
+
+        defer {
+            mutex.unlock()
+        }
+
+        let newOrUpdatedItems = changes.newOrUpdatedItems.filter { delegatedMetaAccount in
+            guard let delegationId = delegatedMetaAccount.info.delegationId() else { return false }
```
ERussel, line 30: I think we are skipping here chain information and type here. For example, i imagine we treat the same 2 proxies in different chains or proxy vs multisig

novawallet/Common/Substrate/Operations/Identity/IdentityProxyFactory.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 import Operation_iOS
 
-protocol IdentityProxyFactoryProtocol {
+protocol IdentityDelegatedAccountFactoryProtocol {
```
ERussel, line 4: It's not related to proxy delegated account but for identity proxy. So i would keep name the same

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChainSyncService.swift
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        return CompoundOperationWrapper(
+            targetOperation: mapOperation,
+            dependencies: fetchWrappers.flatMap(\.allOperations)
+        )
+    }
+
+    func createDiscoverAccountsWrapper(
+        delegatedAccountsSources: [DelegatedAccountsRepositoryProtocol],
```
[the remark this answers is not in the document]
svojsu: added in the following PR

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedAccountsChangesCalculator.swift
```diff
@@ -123,28 +123,22 @@ private extension DelegatedAccountsChangesCalculator {
...
+    ) -> [ManagedMetaAccountModel] {
         let relevantLocalAccounts = localDelegatedAccounts.filter {
             $0.key.delegatorAccountId == delegatorAccountId
-        }
+        }.values
 
-        return relevantLocalAccounts.compactMapValues { metaAccount in
-            guard let factory = getFactoryForMetaAccount(metaAccount) else { return nil }
+        let revokedAccounts: [ManagedMetaAccountModel] = Set(relevantLocalAccounts)
+            .subtracting(Set(updatedDelegatedAccounts))
```
ERussel, line 134: Is it valid to subtract sets of wallets? What if we have wallets with the same id but different data. In the last case such wallet will be different.

novawallet/Common/Helpers/MetaAccountModel+Delegated.swift
```diff
@@ -0,0 +1,119 @@
...
+            )
+        case .proxied:
+            guard
+                let proxyAccountId = proxy()?.accountId,
+                let proxiedAccount = chainAccounts.first(where: { $0.proxy?.accountId == proxyAccountId }),
+                let proxy = proxiedAccount.proxy
+            else { return nil }
+
+            return MetaAccountDelegationId(
+                delegatedAccountId: proxyAccountId,
```
ERussel, line 76: We have discussed that we will have `delegate` and `delegator` notion instead of delegated

novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedAccountsChangesCalculator.swift
svojsu: Yes, you're right. It's better to subtract delegationIds

ERussel (approved):

PR #1525, opened 2025-05-13
Delegated accounts sync refactor

svojsu:
### SUMMARY
The current implementation of `DelegatedAccountSyncService` and `DelegatedAccountChainSyncService` leads to the following issues:
- Universal multisig meta accounts created by `DelegatedAccountChainSyncService` cannot be differentiated when we need to mark a multisig as revoked.
- Meta accounts might be incorrectly removed by `DelegatedAccountSyncService` during subsequent status updates due to being wrongly marked as revoked.
- `DelegatedAccountSyncBarrier` should be cleared at the end of the sync process, but we cannot determine when synchronization is complete.

### SOLUTION
`DelegatedAccountSyncService` now serves as the main observable service executing update wrappers, which are created by `DelegatedAccountFetchOperationFactory`. The factory composes long-running operations from chain-specific wrappers and maps all changes, filtering out duplicates and resolving actual delegate account statuses.
This approach eliminates the need for any barrier implementation.

### TODO
Fix mocks generation


novawallet/Common/Services/DelegatedAccounts/SyncService/Factory/DelegatedAccountFetchOperationFactory.swift
```diff
@@ -0,0 +1,129 @@
...
+
+                return managedMetaAccount
+            }
+
+        let resultUpdates = filteredUpdates.map { managedMetaAccount in
+            guard
+                let delegationId = managedMetaAccount.info.delegationId(),
+                let currentStatus = managedMetaAccount.info.delegatedAccountStatus(),
+                let collectedStatuses = delegateStatusMap[delegationId],
+                collectedStatuses.count > 1
```
ERussel, line 67: Didn't get this. What if we get `.revoke` from 2 chains of 3. In that case we still need to keep the delegated account

ERussel (approved):

PR #1527, opened 2025-05-16
Correct mapping for nested proxies and multisigs

svojsu:
### SUMMARY
This PR corrects mapping and fetching logic for nested proxies and multisigs.


novawallet/Common/Services/DelegatedAccounts/SyncService/DelegatedAccountChangesCalculator/DelegatedMetaAccountFactory/MultisigMetaAccountFactory.swift
```diff
@@ -65,16 +65,16 @@ extension MultisigMetaAccountFactory: DelegatedMetaAccountFactoryProtocol {
...
         guard
             let multisig = delegatedAccount as? DiscoveredMultisig,
             let multisigAccountType = createMultisigType(
                 discoveredMultisig: multisig,
-                localMetaAccounts: localMetaAccounts
+                metaAccounts: metaAccounts
             )
         else {
-            throw DelegatedAccountError.invalidAccountType
+            return nil
```
ERussel, line 77: Why we should return nil here? As I understand if we couldn't create a wallet then it should be an error

svojsu: Yes, you're right. We have discussed it. Fixed.

PR #1528, opened 2025-05-19
Pay tab skeleton & shop brands list [CU &8698uzgy5]

ERussel:
## Purpose

- add PayRoot module with the paging logic: page provider is created by presenter based on the settings interactor derives. PayRoot view controller then manages the switch between pages provided by provider.

- `ScrollDecorationProviding` protocol is added and implemented by PayRoot view controller to allow top bar blurring to go under the segmented control

- PayShop module is introduced with brands loading support. It also implements load more feature. However the state for the unsupported wallet case as well as previous payments and recommended brands will be introduced in the later PRs.

##

https://github.com/user-attachments/assets/9f017242-466d-4ff3-a0a1-10ac3319d114


novawallet/Modules/Pay/Root/PayRootViewLayout.swift
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        barExtendingView = view
+
+        return view
+    }
+}
+
+private extension PayRootViewLayout {
+    private func setupLayout() {
```
svojsu, line 61: access modifier is not necessary

svojsu (approved):

PR #1529, opened 2025-05-20
Multisig sender resolving

svojsu:
### SUMMARY
This PR refactors the proxy sender resolution logic, improving the resolution graph, path merger, path finder, and resolver to work with abstract delegation data.

### DIFF
--- a/novawallet/Common/Storage/WalletsUpdateMediator.swift
+++ b/novawallet/Common/Storage/WalletsUpdateMediator.swift
@@ -38,29 +38,34 @@ final class WalletUpdateMediator {
         self.operationQueue = operationQueue
     }
 
-    private static func nonProxiedWalletReachable(
-        from proxiedWallet: ManagedMetaAccountModel,
+    private static func nonDelegatedWalletReachable(
+        from delegatedWallet: ManagedMetaAccountModel,
         wallets: [ManagedMetaAccountModel],
         removeIds: Set<MetaAccountModel.Id>
     ) -> Bool {
-        var currentProxieds: Set<MetaAccountModel.Id> = [proxiedWallet.info.metaId]
-        var prevProxieds = currentProxieds
-        var foundWallets: [MetaAccountModel.Id: ManagedMetaAccountModel] = [proxiedWallet.info.metaId: proxiedWallet]
+        var currentDelegatedSet: Set<MetaAccountModel.Id> = [delegatedWallet.info.metaId]
+        var prevDelegated = currentDelegatedSet
+        var foundWallets: [MetaAccountModel.Id: ManagedMetaAccountModel] = [delegatedWallet.info.metaId: delegatedWallet]
 
         repeat {
-            let newReachableWallets: [ManagedMetaAccountModel] = currentProxieds.flatMap { proxiedId in
+            let newReachableWallets: [ManagedMetaAccountModel] = currentDelegatedSet.flatMap { delegatedId in
                 guard
-                    let proxied = foundWallets[proxiedId],
-                    let chainAccount = proxied.info.chainAccounts.first(where: { $0.proxy != nil }),
-                    let proxy = chainAccount.proxy else {
+                    let delegated = foundWallets[delegatedId],
+                    let delegationId = delegated.info.delegationId() else {
                     return [ManagedMetaAccountModel]()
                 }
 
-                return wallets.filter { $0.info.has(accountId: proxy.accountId, chainId: chainAccount.chainId) }
+                let filter: (ManagedMetaAccountModel) -> Bool = if let chainId = delegationId.chainId {
+                    { $0.info.has(accountId: delegationId.delegateAccountId, chainId: chainId) }
+                } else {
+                    { $0.info.contains(accountId: delegationId.delegateAccountId) }
+                }
+
+                return wallets.filter(filter)
             }
 
             if newReachableWallets.contains(
-                where: { $0.info.type != .proxied && !removeIds.contains($0.info.metaId) }
+                where: { !$0.info.isDelegated() && !removeIds.contains($0.info.metaId) }
             ) {
                 return true
             }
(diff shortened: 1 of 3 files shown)


novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/DelegationResolutionPathFinder.swift
```diff
@@ -164,3 +179,38 @@ extension ProxyResolution {
...
+        let signatories: [AccountId]
+
+        func wrapCall(
+            _ call: JSON,
+            delegation _: DelegationKey,
+            context _: RuntimeJsonContext
+        ) throws -> JSON {
+            // TODO: - Implement call wrapping
+
+            call
```
ERussel, line 213: So, here i would consider the following approaches (not sure yet which one to pick):

1) Check whether the call is from multisig pallet and intended to be signed by the same signatory as we have resolved. The idea here is to wrap only with as_multi but we need to differentiate from the cases where multisig is signatory of another multisig.

2) Introduce dummy call for approval and rejection and check for them here.

3) Wrap always here with as_multi. And when we perform an operation resolve the signatory in place and use current resolver for already wrapped call.

I would go with 3 for now as it one line for that function and continue the discussion in the approval/rejection operation task.

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/DelegationResolutionGraph.swift
```diff
@@ -0,0 +1,243 @@
...
+    case emptyPaths
+    case disjointPaths
+}
+
+struct DelegationKey: Hashable {
+    let delegate: AccountId
+    let delegated: AccountId
+}
+
+protocol AccountDelegationGraphValue {
```
ERussel, line 17: I would manually emphasize that it is protocol. For example, DelegationResolutionNodeProtocol

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/DelegationResolutionGraph.swift
```diff
@@ -0,0 +1,243 @@
...
+
+enum DelegationError: Error {
+    case noCompatibleProxyType
+    case invalidMultisigParameters
+    case noValidPath
+    case emptyPaths
+    case disjointPaths
+}
+
+struct DelegationKey: Hashable {
```
ERussel, line 12: Probably the error above and this struct should be defined inside DelegationResolution namespace

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/DelegationResolutionPathFinder.swift
```diff
@@ -1,6 +1,15 @@
 import Foundation
+import SubstrateSdk
+
+protocol AccountDelegationPathValue {
```
ERussel, line 4: `DelegationResolutionPathNodeProtocol` maybe?

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/DelegationResolutionPathFinder.swift
```diff
@@ -141,13 +156,13 @@ extension ProxyResolution {
...
-            from paths: [CallCodingPath: ProxyResolution.GraphResult]
-        ) throws -> ProxyResolution.PathFinderResult {
+            from paths: [CallCodingPath: DelegationResolution.GraphResult]
+        ) throws -> DelegationResolution.PathFinderResult {
             if let secretBasedResult = find(callPaths: paths, walletTypeFilter: { $0 == .secrets }) {
                 return secretBasedResult
             } else if let notWatchOnlyResult = find(
                 callPaths: paths,
-                walletTypeFilter: { $0 != .watchOnly && $0 != .proxied }
+                walletTypeFilter: { $0 != .watchOnly && $0 != .proxied && $0 != .multisig }
```
ERussel, line 165: I think it will be got to have calculated property for the wallet type - `isDelegated`

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/Delegation/ExtrinsicDelegateSenderResolver.swift
```diff
@@ -31,33 +31,35 @@ final class ExtrinsicProxySenderResolver {
...
-                guard let proxyPath = solution.callToPath[callPath] else {
+                guard let delegatePath = solution.callToPath[callPath] else {
                     return callJson
                 }
 
-                let (resultCall, _) = try proxyPath.components.reduce(
-                    (callJson, proxiedAccount.accountId)
+                let (resultCall, _) = try delegatePath.components.reduce(
+                    (callJson, delegatedAccount.accountId)
                 ) { callAndProxied, component in
```
ERussel, line 50: callAndDelegatedAccount

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/ExtrinsicSenderResolutionFactory.swift
```diff
@@ -63,10 +67,20 @@ extension ExtrinsicSenderResolutionFactory: ExtrinsicSenderResolutionFactoryProt
     func createWrapper() -> CompoundOperationWrapper<ExtrinsicSenderResolving> {
         switch chainAccount.type {
         // TODO: Add multisig resolver
```
ERussel, line 69: Do we need the TODO now?

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/ExtrinsicSenderResolutionFactory.swift
```diff
@@ -63,10 +67,20 @@ extension ExtrinsicSenderResolutionFactory: ExtrinsicSenderResolutionFactoryProt
...
-        case .secrets, .paritySigner, .polkadotVault, .ledger, .watchOnly, .genericLedger, .multisig:
-            return createCurrentResolver(for: chainAccount)
+        case .secrets, .paritySigner, .polkadotVault, .ledger, .watchOnly, .genericLedger:
+            createCurrentResolver(for: chainAccount)
         case .proxied:
```
ERussel, line 72: This probably could be simpler if we have just `delegatedAccountId` instead of delegateKeyPath (which seems should to be delegatedKeyPath). As our MetaAccountModel already has such property we can combine `case .proxied, .multisig:`.

novawallet/Common/Storage/WalletsUpdateMediator.swift
```diff
@@ -135,7 +135,7 @@ final class WalletUpdateMediator {
             // we want to change selected wallet if current one is removed or revoked as proxied
 
             let newUpdatedWallets = changes.newOrUpdatedItems.map { wallet in
-                if wallet.isSelected, wallet.info.proxy()?.status == .revoked {
+                if wallet.isSelected, wallet.info.proxy?.status == .revoked {
```
ERussel, line 138: Just noticed that we didn't move `WalletsUpdateMediator` to delegation notion yet but it should be done in terms of delegated account detection logic

ERussel (approved):

novawallet/Common/Services/ExtrinsicService/Substrate/ExtrinsicSenderResolution/ExtrinsicSenderResolutionFactory.swift
svojsu: I don't think we're able to pass `delegatedAccountId` without butchering the method. The id is received from wallets that need to be fetched from the repository. Currently, the method encapsulates the logic very well.

Actually, we can pass a single keypath to `delegateAccountId` so it will be resolved implicitly. I've added this solution.

Regarding `delegateKeyPath` vs `delegatedKeyPath`, it was a typo in the multisig case. We are aiming to receive the delegateAccountId, so it needed to be a keypath to the signatory.

novawallet/Common/Helpers/MetaAccountModel+Delegated.swift
```diff
@@ -93,6 +62,36 @@ extension MetaAccountModel {
...
+        type.isDelegated
+    }
+
+    func proxyChainAccount(
+        chainId: ChainModel.Id
+    ) -> ChainAccountModel? {
+        chainAccounts.first { $0.chainId == chainId && $0.proxy != nil }
+    }
+
+    func address(for chainAsset: ChainAsset) throws -> AccountAddress? {
```
ERussel, line 90: Look like having chainAsset here to much. WDYT to leave only chain?

ERussel (approved):

PR #1537, opened 2025-05-26
Feature/sign multisig transaction

svojsu:
### SUMMARY

This PR adds wrapping for the call in an AsMulti call and refactors the signer UI logic to compose the multisig flow with the proxy flow.


novawallet/Common/Substrate/Calls/Multisig/Multisig+Call.swift
```diff
@@ -0,0 +1,53 @@
...
+        enum CodingKeys: String, CodingKey {
+            case threshold
+            case otherSignatories = "other_signatories"
+            case maybeTimepoint = "maybe_timepoint"
+            case call
+            case maxWeight = "max_weight"
+        }
+
+        @StringCodable var threshold: UInt16
+        let otherSignatories: [AccountId]
```
ERussel, line 16: It should be a list of `BytesCodable` otherwise it will not work on some networks like evm

novawallet/Common/Substrate/Calls/Multisig/Multisig+Call.swift
```diff
@@ -0,0 +1,53 @@
...
+        @NullCodable var maybeTimepoint: MultisigTimepoint?
```
ERussel, line 17: I don't see any need to use `NullCodable`

novawallet/Common/Substrate/Calls/Multisig/Multisig+Call.swift
```diff
@@ -0,0 +1,53 @@
...
+        let call: JSON
+        let maxWeight: Weight
```
ERussel, line 19: Check that in the end it is v2 Weight

svojsu: It's better to make the struct non-generic then.

PR #1549, opened 2025-06-02
Don't persist multisig transfer extrinsic

svojsu:
### SUMMARY
The PR adds filter logic for extrinsics persisted in transfer confirm interactors.


novawallet/Modules/Transfer/BaseTransfer/OnChain/OnChainTransferBaseInteractor.swift
```diff
@@ -16,6 +16,7 @@ class OnChainTransferBaseInteractor {
     let asset: AssetModel
     let walletLocalSubscriptionFactory: WalletLocalSubscriptionFactoryProtocol
     let priceLocalSubscriptionFactory: PriceProviderFactoryProtocol
+    let persistenceFilter: ExtrinsicPersistenceFilterProtocol
```
ERussel, line 19: Why do we need the dependency in the base class if there is no usage of it?

svojsu: Actually we don't. Fixed.

ERussel (approved):

PR #1581, opened 2025-07-03
Feature/check feature availability for msig wallet

svojsu:
## Purpose

Introduce `FeatureSupportChecking` to perform availabillity of the operation for a given wallet. Replace validation for buy/sell with the new approach

Closes &8699dgnv6 &8699dgnv6 &8699hh97m &8699gz5cv &8699hdf35


novawallet/Common/Protocols/FeatureSupportChecking.swift
```diff
@@ -0,0 +1,100 @@
...
+        let unsupportedClosure: () -> Void = { [weak self] in
+            self?.presentFeatureUnsupportedView(
+                from: sheetPresentingView,
+                type: .sell,
+                walletType: .init(walletType: multisigMetaAccount.type),
+                completion: {}
+            )
+        }
+
+        checkIfMultisigThreshold1(
```
ERussel, line 29: This approach will not work for complex account structure - when current wallet is proxy and there is a multisig somewhere in hierarchy. We need integrate `WalletDelayedExecVerifier` for that.

PR #1587, opened 2025-07-04
Implement call formatting logic

ERussel:
## Purpose

PR introduces `CallFormattingOperationFactory` that allows async call formatting. Currently it is supported either general call (module + call name) and transfers. The implementation is based on refactored `NestedCallMapper` that allows to get leaf call also returning the actual origin of it. The origin is used to display the account on behalf the transaction is executed.

The second part of the PR integrates the formatter into the pending operation subscription logic introducing `MultisigOperationProviderProxy`. The proxy enriches the pending operation change recevied from the subscription with the formatted model and passes further to interactors.


novawallet/Modules/DelegatedAccounts/Multisig/PendingOperationProxy/MultisigPendingOperationProviderProxy.swift
```diff
@@ -0,0 +1,263 @@
+import Foundation
+import Operation_iOS
+
+protocol MultisigOperationProviderProxyProtocol: AnyObject {
+    var handler: MultisigOperationProviderHandlerProtocol? { get set }
```
svojsu, line 5: Why would we want it as get-set property of the interface instead of passing the handler into subscription methods?

svojsu (approved):

PR #1590, opened 2025-07-05
Multisig transaction full details

svojsu:
### SUMMARY

PR introduces `MultisigTxDetails` module and extends `MultisigDefinition` model with `deposit` field.

**UPD.** Fixed Network Fee view leading inset
**UPD.** Fixed wrong text on call hash actions

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/a3e2d6ba-5a97-4f59-bfa4-1d63749a7d09


novawallet/Common/Extension/Foundation/String+Truncation.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+extension String {
+    func truncatedMiddle(
```
ERussel, line 4: We have the same logic in `extension AccountAddress`. I would rather rename it to `extension String` as AccountAddress is a type alias to String

novawallet/Common/Substrate/Calls/Builder/RuntimeCall+JSON.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import SubstrateSdk
+
+extension RuntimeCall where T == JSON {
+    func toCallJSON() -> JSON {
```
ERussel, line 5: WDYT aboun `toDisplayRepresentation`?

novawallet/Modules/DelegatedAccounts/Multisig/MultisigTxDetails/Factory/MultisigTxDetailsViewModelFactory.swift
```diff
@@ -0,0 +1,229 @@
...
+
+    func createDepositViewModel(
+        deposit: BigUInt,
+        depositAsset: ChainAsset,
+        assetPrice: PriceData?,
+        locale: Locale
+    ) -> BalanceViewModelProtocol {
+        let assetInfo = depositAsset.asset.displayInfo
+
+        let depositDecimal = Decimal.fromSubstrateAmount(
```
ERussel, line 161: `deposit.decimal(assetInfo: assetInfo)`. Could you, please, also refactor other places in the file if there is any

ERussel (approved):

PR #1613, opened 2025-07-28
Add MS operation reject confirmation sheet

svojsu:
### SUMMARY

The PR adds reject confirmation screen for multisig operations

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/88001f2b-db89-45fb-99d2-33caa5c9285d

### DIFF
--- a/novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationConfirm/MultisigOperationConfirmPresenter.swift
+++ b/novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationConfirm/MultisigOperationConfirmPresenter.swift
@@ -12,6 +12,7 @@ final class MultisigOperationConfirmPresenter {
     let chain: ChainModel
     let multisigWallet: MetaAccountModel
 
+    var needsConfirmation: Bool?
     var signatories: [Multisig.Signatory]?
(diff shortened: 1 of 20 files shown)


novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationConfirm/MultisigOperationConfirmPresenter.swift
```diff
@@ -173,6 +174,36 @@ private extension MultisigOperationConfirmPresenter {
...
     func doConfirm() {
+        validateConfirmation { [weak self] in
+            guard let self else { return }
+
+            let confirmClosure = {
+                self.view?.didReceive(loading: true)
+                self.interactor.confirm()
+            }
+
+            if let needsConfirmation, needsConfirmation {
```
sergkononenko, line 185: Wouldn't it be more readable to use 
```
let needsConfirmation, needsConfirmation else {
confirmClosure()
return
} 
```
and then an additional `guard` ?

svojsu: Yeah, you're right. Fixed.

PR #1641, opened 2025-08-19
Feature/multisig notifications formatting

svojsu:
### SUMMARY
The PR introduces multisig push notification handling and presentation implementation.

### SOLUTION
can now be constructed to work with either persisted chains and metadata or a full-blown network-dependent registry and providers. To achieve this behavior, the `chainRegistry` dependency has been separated into `chainProvider` and `runtimeCodingServiceProvider`.

**RuntimeCodingServiceProvider**

`ChainRegistry` implements `RuntimeCodingServiceProviderProtocol` and provides the full runtime service implementation for the novawallet target.

`OfflineRuntimeCodingService` implements `RuntimeCodingServiceProviderProtocol` and provides an offline runtime coding service implementation for the notifications extension target.

**ChainProvider**

`ChainRegistry` implements `ChainProviderProtocol` and provides chains from the registry for the novawallet target.

`OfflineChainProvider` implements `ChainProviderProtocol` and provides chains based on persistence for the notifications extension target.

This approach enables us to construct and use `CallFormattingOperationFactory` for both novawallet and notifications extension targets.

### SCREENSHOTS

IMG_5981
IMG_5982


NovaPushNotificationServiceExtension/Handlers/Multisig/MultisigApprovalHandler.swift
```diff
@@ -0,0 +1,14 @@
...
+final class MultisigApprovalHandler: CommonMultisigHandler, PushNotificationHandler {
+    override func createTitle(using payload: MultisigPayloadProtocol) -> String {
+        R.string.localizable.pushNotificationMultisigApprovalTitle(
+            payload.signatoryAddress.mediumTruncated,
+            preferredLanguages: locale.rLanguages
+        )
+    }
+
+    override func createBody(using _: MultisigPayloadProtocol) -> String {
+        ""
```
ERussel, line 12: Shouldn't we display information about call in the body? On the screenshots i see title with "Signed by ..." and body "Uknown operation on Kusama", for example

novawallet/Common/Extension/Foundation/String+Substrate.swift
```diff
@@ -29,9 +29,11 @@ extension String {
     }
 
     func replacingCamelCase() -> String {
-        var replacedCamelCase: String = ""
+        guard let first else { return self }
+
+        var replacedCamelCase: String = "\(first.lowercased())"
         let upperCase = CharacterSet.uppercaseLetters
-        for scalar in unicodeScalars {
+        for scalar in unicodeScalars.dropFirst() {
```
ERussel, line 36: Could you, please, give an example what does this fix?

novawallet/Common/Services/CallFormatting/CallFormattingOperationFactory.swift
```diff
@@ -225,48 +233,104 @@ private extension CallFormattingOperationFactory {
             decoded: decodedCall
         )
     }
+
+    func createDecodingWrapper(
+        dependingOn runtimeCodingServiceWrapper: CompoundOperationWrapper<RuntimeCodingServiceProtocol>,
+        for callData: Substrate.CallData,
+        chainId _: ChainModel.Id
```
ERussel: Looks like unused

novawallet/Common/Services/CallFormatting/CallFormattingOperationFactory.swift
```diff
@@ -225,48 +233,104 @@ private extension CallFormattingOperationFactory {
...
+                of: GenericType.call.name
+            )
+        }
+
+        return decodingWrapper
+    }
+
+    func createCodingFactoryWrapper(
+        dependingOn runtimeCodingServiceWrapper: CompoundOperationWrapper<RuntimeCodingServiceProtocol>,
+        chainId _: ChainModel.Id
```
ERussel: Looks like unused

novawallet/Common/Services/CallFormatting/CallFormattingOperationFactory.swift
```diff
@@ -225,48 +233,104 @@ private extension CallFormattingOperationFactory {
...
+            return codingService.createDecodingWrapper(
+                for: callData,
+                of: GenericType.call.name
+            )
+        }
+
+        return decodingWrapper
+    }
+
+    func createCodingFactoryWrapper(
```
ERussel, line 255: I think we could just a add to the extension of RuntimeCodingServiceProviderProtocol a method of getting wrapper with RuntimeCoderFactoryProtocol value directly

novawallet/Modules/Notifications/NotificationHandlers/MultisigPayload/MultisigPayload.swift
```diff
@@ -0,0 +1,45 @@
...
+    var callData: Substrate.CallData? { get }
+}
+
+struct MultisigPayload<Keys: MultisigCodingKeyProtocol>: MultisigPayloadProtocol {
+    let multisigAddress: AccountAddress
+    let signatoryAddress: AccountAddress
+    @HexCodable var callHash: Substrate.CallHash
+    @OptionHexCodable var callData: Substrate.CallData?
+
+    init(from decoder: Decoder) throws {
```
ERussel, line 23: Why do we need to implement manual encoding/decoding here?

novawallet/Common/Substrate/Types/Xcm/XcmPalletMetadataQueryFactory.swift
```diff
@@ -5,23 +5,23 @@ import BigInt
 
 protocol XcmPalletMetadataQueryFactoryProtocol {
     func createModuleNameResolutionWrapper(
-        for runtimeProvider: RuntimeProviderProtocol
```
ERussel, line 8: What is the purpose of changing types here?

Podfile
```diff
@@ -82,6 +82,7 @@ abstract_target 'novawalletAll' do
     pod 'sr25519c', :git => 'https://github.com/novasamatech/sr25519.c', :tag => '0.2.0'
     pod 'ed25519.c', :git => 'https://github.com/novasamatech/ed25519.c', :tag => '0.1.1'
     pod 'blake2.c', :git => 'https://github.com/novasamatech/blake2.c', :tag => '0.1.0'
+    pod 'web3swift', :git => 'https://github.com/web3swift-team/web3swift.git', :tag => '3.0.6'
```
ERussel, line 85: Why do we need this dependency in the extension?

novawallet/Common/Services/ChainProvider/ChainRegistry+ChainProvider.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import Operation_iOS
+
+extension ChainRegistry: ChainProviderProtocol {
```
ERussel, line 4: I would just do ChainRegistryChainProvider that has chain registry protocol inside. In that case we don't need to case ChainRegistryProtocol to ChainRegistry below. Also it will be consistent with `OfflineChainProvider` approach

novawallet/Modules/Notifications/NotificationHandlers/MultisigPayload/MultisigPayload+Rejected.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+extension MultisigPayloadCodingKey {
```
ERussel, line 3: I was hard to understand how this work as generic parameter for the model. I think it is more readable to create separate models for each payload type. WDYT?

NovaPushNotificationServiceExtension/Handlers/Multisig/MultisigApprovalHandler.swift
svojsu: It's specific body part for approval notification that should have "approvals of threshold" value and currently is not available. Common part is still presented via superclass logic.

novawallet/Common/Extension/Foundation/String+Substrate.swift
svojsu: Because the approach is to create spaces before each capital, any string started from the capital letter resulted in space before the string. You can see the problem on the screenshots below.

Before:




After:

novawallet/Common/Services/CallFormatting/CallFormattingOperationFactory.swift
svojsu: fixed

novawallet/Modules/Notifications/NotificationHandlers/MultisigPayload/MultisigPayload.swift
svojsu: separated payload DTOs for multisig

Podfile
svojsu: Temporary addition until I sort out imports for files included in extension target

novawallet/Common/Services/ChainProvider/ChainRegistry+ChainProvider.swift
svojsu: Did it for both `ChainProvider` and `RuntimeCodingServiceProvider`

PR #1645, opened 2025-08-20
Feature/multisig notifications actions

svojsu:
### SUMMARY
This PR:
1. Introduces handling logic for Multisig notifications.
2. Fixes an old issue with Multisig wallet naming. Currently, addresses are formatted based on the chain they were discovered on.

### SOLUTION

**Notifications handling**

Actions are handled in the same manner as previous notifications like transfers or staking rewards. The key consideration is how the `MultisigOperationConfirm` module is opened. The current implementation assumes we have a local operation, so we can determine which interactor (reject or approve) to use while constructing the module. 

`MultisigOperationFetchProxy` module is introduced to comply with this implementation. It's a proxy module that takes an operation key as input. It then fetches an on-chain operations and looks for desired `callHash`. If the operation is not found, it presents corresponding alert. Otherwise, it persists the operation and passes directly to `MultisigOperationConfirm`.

The action handling logic of `MultisigNotificationMessageHandler` requires wallet switching. Since this logic is the same as `AssetDetailsNotificationMessageHandler`, we introduce a `WalletSelectingNotificationHandling` trait. The trait is self-contained and takes care of finding and selecting the right wallet. Both `MultisigNotificationMessageHandler` and `AssetDetailsNotificationMessageHandler` support this trait instead of implementing the switching logic individually.

**Multisig naming fix**

To comply with the Android approach, we format the substrate address with a 0-prefix. A `toAddressWithDefaultConversion` method is introduced for `AccountId` to create the address with conversion depending on the address scheme. It relies on the address length when deciding which scheme to use.

### SCREENSHOTS

IMG_5999
IMG_5994
IMG_6001

### SCREEN RECORDINGS

https://github.com/user-attachments/assets/a61a0da4-ba94-4512-9ac3-f90714007c57


novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationConfirm/Factory/MultisigOperationConfirmViewFactory.swift
```diff
@@ -5,7 +5,7 @@ import Foundation_iOS
 struct MultisigOperationConfirmViewFactory {
     static func createView(
         for operation: Multisig.PendingOperationProxyModel,
-        flowState: MultisigOperationsFlowState
+        flowState: MultisigOperationsFlowState?
```
ERussel, line 8: Is there any purpose to make it optional?

novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationFetchProxy/MultisigOperationFetchProxyWireframe.swift
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    func showConfirmationData(
+        from view: ControllerBackedProtocol?,
+        for operation: Multisig.PendingOperationProxyModel,
+        flowState: MultisigOperationsFlowState
+    ) {
+        guard let confirmView = MultisigOperationConfirmViewFactory.createView(
+            for: operation,
+            flowState: flowState
```
ERussel, line 25: It looks weird to make a trip interactor -> presenter -> wireframe for the flow state. I would rather create the state directly in the view factory, store `MultisigOperationProviderProxy` in it and pass state further. Then we will have shared provider for both operations and operation details. WDYT?

novawallet/Modules/DelegatedAccounts/Multisig/MultisigOperationConfirm/Factory/MultisigOperationConfirmViewFactory.swift
svojsu: It was one of the iterations. Fixed.

ERussel (approved):

PR #1656, opened 2025-08-25
Fix/push notification content

svojsu:
### SUMMARY

The PR adds logic to find and present wallet name if it's persisted or address if it's not.
It also adds approvals count for approval notifications content.

### DIFF
--- a/NovaPushNotificationServiceExtension/Handlers/Multisig/CommonMultisigHandler.swift
+++ b/NovaPushNotificationServiceExtension/Handlers/Multisig/CommonMultisigHandler.swift
@@ -104,11 +104,7 @@ private extension CommonMultisigHandler {
         _ address: String,
         for chain: ChainModel
     ) -> AccountAddress {
-        (
-            try? address
-                .toChainAccountIdOrSubstrateGeneric(using: chain.chainFormat)
-                .toAddress(using: chain.chainFormat)
-        ) ?? address
+        (try? address.toAccountId().toAddress(using: chain.chainFormat)) ?? address
     }
 
     func createNotificationParams(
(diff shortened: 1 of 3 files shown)


NovaPushNotificationServiceExtension/Handlers/Multisig/CommonMultisigHandler.swift
```diff
@@ -102,44 +100,77 @@ extension CommonMultisigHandler {
...
 private extension CommonMultisigHandler {
-    func createNotificationContentWrapper(
-        wallets: [Web3Alert.LocalWallet],
+    func convertAddress(
+        _ address: String,
+        for chain: ChainModel
+    ) -> AccountAddress {
+        (
+            try? address
+                .toChainAccountIdOrSubstrateGeneric(using: chain.chainFormat)
```
ERussel, line 109: is it important to require that the address matches chain/concrete prefix? Also does evm addresses work here? Probably just use `toAccountId()`

NovaPushNotificationServiceExtension/Handlers/Multisig/NewMultisigHandler.swift
```diff
@@ -1,15 +1,18 @@
...
-    override func createTitle(using _: MultisigPayloadProtocol) -> String {
+    override func createTitle(params _: MultisigNotificationParams) -> String {
         R.string.localizable.pushNotificationMultisigNewTitle(
             preferredLanguages: locale.rLanguages
         )
     }
 
-    override func createBody(using payload: MultisigPayloadProtocol) -> String {
+    override func createBody(
+        using _: MultisigPayloadProtocol,
```
ERussel, line 11: Whether we still need payload here?

ERussel (approved):

PR #1669, opened 2025-09-08
[AH-Migration] Aggregate staking rewards APIs

svojsu:
### SUMMARY

The PR adds support for `staking-rewards` external API list. The values received from the indexers are merged in result value.
`staking-rewards` API is used in place of old `staking` API for rewards and payouts fetch.

### DIFF
--- a/novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
+++ b/novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
@@ -72,6 +72,28 @@ protocol NPoolsLocalSubscriptionFactoryProtocol {
 
 final class NPoolsLocalSubscriptionFactory: SubstrateLocalSubscriptionFactory {}
 
+private extension NPoolsLocalSubscriptionFactory {
+    func createHash(
+        for address: AccountAddress,
+        startTimestamp: Int64?,
+        endTimestamp: Int64?,
+        api: Set<LocalChainExternalApi>
+    ) -> Int {
+        var hasher = Hasher()
+
+        hasher.combine(address)
+        hasher.combine(startTimestamp ?? 0)
+        hasher.combine(endTimestamp ?? 0)
+
+        api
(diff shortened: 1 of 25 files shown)


novawallet/Common/DataProvider/StakingAnalyticsLocalSubscriptionFactory.swift
```diff
@@ -31,11 +31,11 @@ final class StakingAnalyticsLocalSubscriptionFactory {
 extension StakingAnalyticsLocalSubscriptionFactory: StakingAnalyticsLocalSubscriptionFactoryProtocol {
     func getWeaklyAnalyticsProvider(
         for address: AccountAddress,
-        url: URL
+        urls: [URL]
     ) -> AnySingleValueProvider<[SubqueryRewardItemData]> {
         clearIfNeeded()
 
-        let identifier = "weaklyAnalytics" + address + url.absoluteString
+        let identifier = "weaklyAnalytics" + address + urls.map(\.absoluteString).joined(with: .dash)
```
ERussel, line 38: Those identifires could be quite long. I would take hash at the end.

novawallet/Common/DataProvider/NPoolsLocalSubscriptionFactory.swift
```diff
@@ -268,7 +268,9 @@ extension NPoolsLocalSubscriptionFactory: NPoolsLocalSubscriptionFactoryProtocol
             endTimestamp.map { "\($0)" } ?? "nil"
         ].joined(separator: "-")
 
-        let identifier = ("poolReward" + api.url.absoluteString) + address + timeIdentifier
+        let identifier = ("poolReward" + api.map(\.url.absoluteString).joined(with: .dash))
```
ERussel, line 271: Let's hash the identifier. Also i think we need to sort apis by url since we have a set as parameter and it can give different order for the same set

ERussel (approved):

PR #1670, opened 2025-09-08
[AH-Migration] Use timestamp for chain with separate timeline chain

svojsu:
### SUMMARY

This PR adds timestamp usage for the "last period votes count" case. The timestamp is used for indexer query filters when the chain has a separate timeline chain.

### DIFF
--- a/novawallet/Common/Services/TimepointThresholdService.swift
+++ b/novawallet/Common/Services/TimepointThresholdService.swift
@@ -5,6 +5,7 @@ protocol TimepointThresholdServiceProtocol: AnyObject {
     func add(
         observer: AnyObject,
         sendStateOnSubscription: Bool,
+        queue: DispatchQueue?,
         closure: @escaping Observable<TimepointThreshold?>.StateChangeClosure
     )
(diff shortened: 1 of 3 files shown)


novawallet/Common/Substrate/BlockTime/ChainTimelineFacade.swift
```diff
@@ -41,4 +42,36 @@ extension ChainTimelineFacade: ChainTimelineFacadeProtocol {
             return .createWithError(error)
         }
     }
+
+    func createTimepointThreshold(backIn days: Int) -> CompoundOperationWrapper<TimepointThreshold?> {
```
ERussel, line 46: I wouldn't complicate more timeline facade and revert BlockTimeEstimationService. I think the problem is that we need a way to subscribe to a threshold and fetch data based on it once on change. I think we could create a small service inheriting from `BaseObservableStateStore`. It decides based on chain whether to subscribe to a block number or create threshold immediately based on timestamp, then save it to the state. Interactors can start the service in setup and subscribe to it's state.

PR #1671, opened 2025-09-08
Replace CocoaPods with SPM

ERussel:
## Purpose

#869aaj6cq

- Cocoapods support is fully removed
- All needed libraries are added via SPM
- Added Mintfile that allows to bootstrap SwiftLint, SwiftFormat and Sourcery for given versions. The main advantages that there is no limitation in arguments counts as XCode plugin have and version pining comparing to `brew` approach. However there is a limitation of the Mint - it might be that after MacOS major upgrade old versions can't be build as the internal build process hardcodes macossdk.
- Added Brewfile to bootstrap needed dependencies: Mint and Gem. Generamba still should be installed separately - added a note to the README.

- scripts for secrets injection, linting and formating are moved to the .sh files for better tracking. In the build phase we now only have execution of the .sh files.

- Rswift is bumped to 7.8 as Mint can't build old versions due to MacOS compatability.

Previous: `R.string.localizable.commonCancel(preferredLanguages: locale.rLanguages)`
Now: `R.string(preferredLanguages: locale.rLanguages).localizable.commonCancel()`

Note: I have used a script for that conversion and it has been applied in the "fix linting" commits for more carefull check.

- Cuckoo version is bumped as old version had poor support of SPM. Now there is no option to automatically create mocks for protocol/classes from 3rd party libraries. The approach i have used for JSONRPCEngine (from SubstrateSdk):

1. Create test class with dumb protocol implemetation in the test target
2. Add the created class to the Cuckoo's mock config

- We don't use FireMock anymore since it doesn't support SPM. I have just removed the corresponding tests since they only testing conversion from network Data to local models.

- Updated README to reflect the process of dependencies installation and app build & run

Note: CI is still broken and will be fixed a separate PR.


stepanLav:
Task linked: [CU-869aaj6cq iOS. Swift Package Manager Integration](https://app.clickup.com/t/869aaj6cq)

svojsu (approved):

PR #1718, opened 2025-10-05
Prevent logging full list of nominators

ERussel:
## Purpose

The PR fixes an issue when a long list of nominators (#869aq346g) as attempted to be transformed into the string leading to a crash by making sure to print only staking recommendation type and common parameters


stepanLav:
Task linked: [CU-869aq346g Crash on direct staking after "not enough asset to pay fee" error](https://app.clickup.com/t/869aq346g)

svojsu (approved):

PR #1797, opened 2026-01-08
Feature/parachain staking storage update

svojsu:
### SUMMARY
This PR adds support for the `parachain-staking` pallet's `DelegationScheduledRequests` storage following a recent update. Moonriver has already received this update, and Moonbeam will receive it by January 28.

### SOLUTION
Introduced the `ParaStkScheduledRequestsQueryWrapperFactory` class to encapsulate storage determination logic. The storage type is determined by checking the storage metadata's `StorageEntryType`.

The `ScheduledRequest`'s delegator field is now optional, as the new pallet version removed it. The field is preserved for backward compatibility with parachains that may still be using the older pallet version, i.e. Moonbeam or Zeitgeist.


ERussel (approved):
Looks great!

PR #1801, opened 2026-01-15
[NW-2033] Remove non-tradable segments for stablepool exchanges

svojsu:
## SUMMARY

The PR extends `AssetsHydraStableswapExchange.availableDirectSwapConnections` with tradability check for each edge. If check is failed, we skip the edge.

### DIFF
--- a/novawallet/Common/Services/AssetExchange/HydraExchange/Stableswap/AssetsHydraStableswapExchange.swift
+++ b/novawallet/Common/Services/AssetExchange/HydraExchange/Stableswap/AssetsHydraStableswapExchange.swift
@@ -4,6 +4,7 @@ import Operation_iOS
 final class AssetsHydraStableswapExchange {
     let swapFactory: HydraStableswapTokensFactory
     let quoteFactory: HydraStableswapQuoteFactory
+    let tradabilityFactory: HydraStableswapTradabilityFactory
     let host: HydraExchangeHostProtocol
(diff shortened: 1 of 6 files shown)


novawallet/Common/Services/AssetExchange/HydraExchange/Stableswap/AssetsHydraStableswapExchange.swift
```diff
@@ -4,28 +4,50 @@ import Operation_iOS
...
+        assetOut: HydraDx.AssetId,
+        tradabilities: [HydraStableswap.TradabilityPairKey: HydraStableswap.Tradability]
+    ) -> Bool {
+        let tradabilityKey = HydraStableswap.TradabilityPairKey(
+            assetIn: assetIn,
+            assetOut: assetOut
+        )
+
+        return if let pairTradability = tradabilities[tradabilityKey] {
+            pairTradability.canBuy() && pairTradability.canSell()
```
ERussel, line 36: Am i got it correctly that we only allow an edge from assetIn to assetOut if we can both exchange assetIn to assetOut and assetOut to assetIn. It probably worth to check if there are cases where it still make sense to return an edge even if it is not possible to swap from assetOut to assetIn. As we fixed the typo in the worst case the use should face the quote calc error. But i don't think it is blocking the fix

ERussel (approved):

PR #1806, opened 2026-01-26
AppStore Migration Info

svojsu:
## SUMMARY

The PR introduces InfoPopup module to present the same styled pop-up info screens with different content and business logic. 

It also introduces `ASMInfoPopup` module presenting app store migration info and entry point.

It also removes `AHMInfo` module in favor of `AHMInfoPopup`. The new module is basically the same thing, but works on the rails of `InfoPopup` base module.

### SCREENSHOTS

Note: ASM banner is not ready yet on utils side, but the module works with whatever banner path passed.

Simulator Screenshot - iPhone 17 Pro Max - 2026-01-27 at 11 33 38
Simulator Screenshot - iPhone 17 Pro Max - 2026-01-27 at 09 51 13


svojsu:
Deprioritized

PR #1815, opened 2026-02-09
10.8.1 Watch-only updates and Liquid Glass UI

svojsu:


svojsu:
Release severity: Normal
Release version: 10.8.1
Release notes:

In-app warnings to protect users from scams in Nova Wallet 10.8.1!

- Added in-app warnings to help protect users from scams involving Watch-only wallets
- Added the ability to create a Watch-only wallet using only an EVM address
- Added support for Liquid Glass UI elements in iOS 26
- Added support for Hydration slip fee update
- Fixed an issue with routing for some Hydration swaps

ERussel (approved):

PR #1824, opened 2026-04-08
Add Nova service fee for Hydration swaps

Leemo94:
⚠️ Requires ToS update

Introduce a 0.85% commission on Hydration swap output, calculated from the post-slippage minimum amount and transferred to a Nova fee account via a batched extrinsic (batchAll with swap + transfer).

- Generic BundleExtraActions hook on AtomicSwapOperation keeps the AssetExchange layer fee-agnostic; fee logic lives in SwapService
- Rate display bakes in the fee (MetaMask-style UX)
- Conditional "Includes 0.85% Nova fee." disclaimer on swap screens
- Rate tooltip is fee-aware for Hydration swaps, generic otherwise
- "Network fees" label replaces "Total fee"
- Price impact intentionally excludes the commission
- Cross-chain swaps: commission applies to the Hydration segment only
- NovaFeeTransferFilter hides fee transfers from transaction history
- NovaSwapCommissionClosureFactory with resolveCommission() helper
- AssetExchangeQuote+NovaFee extension (SRP-compliant)
- Fee transferred via Tokens/Currencies.transfer for ORML tokens, Balances.transferKeepAlive for native HDX
- 13 language translations for fee-related UI strings

### DIFF
--- a/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
+++ b/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
@@ -153,12 +153,20 @@ final class SwapSetupPresenter: SwapBasePresenter {
             }
 
             payAmountInput = payAmount.map { .absolute($0) }
+
+            if quote.willCollectCommission {
+                receiveAmountInput = receiveChainAsset.map {
+                    quote.displayAmountOut.decimal(assetInfo: $0.asset.displayInfo)
+                }
+                provideReceiveAmountInputViewModel()
+            }
+
             providePayAmountInputViewModel()
             providePayInputPriceViewModel()
             provideReceiveInputPriceViewModel()
         case .sell:
             receiveAmountInput = receiveChainAsset.map {
-                quote.route.quote.decimal(assetInfo: $0.asset.displayInfo)
+                quote.displayAmountOut.decimal(assetInfo: $0.asset.displayInfo)
             }
 
             provideReceiveAmountInputViewModel()
@@ -172,6 +180,8 @@ final class SwapSetupPresenter: SwapBasePresenter {
         provideButtonState()
         provideDetailsViewModel()
         estimateFee()
+
+        view?.didReceiveNovaFeeDisclaimer(visible: quote.willCollectCommission)
     }
 
     override func handleNewFee(
@@ -449,7 +459,7 @@ extension SwapSetupPresenter {
                 assetDisplayInfoIn: assetDisplayInfoIn,
                 assetDisplayInfoOut: assetDisplayInfoOut,
                 amountIn: quote.route.amountIn,
-                amountOut: quote.route.amountOut
+                amountOut: quote.displayAmountOut
             ),
             locale: selectedLocale
         )
@@ -791,7 +801,7 @@ extension SwapSetupPresenter: SwapSetupPresenterProtocol {
     }
 
     func showRateInfo() {
-        wireframe.showRateInfo(from: view)
+        wireframe.showRateInfo(from: view, includesNovaFee: quote?.willCollectCommission ?? false)
     }
 
     func showRouteDetails() {
--- a/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
+++ b/novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
@@ -156,7 +156,7 @@ final class SwapSetupPresenter: SwapBasePresenter {
 
             if quote.willCollectCommission {
                 receiveAmountInput = receiveChainAsset.map {
-                    quote.displayAmountOut.decimal(assetInfo: $0.asset.displayInfo)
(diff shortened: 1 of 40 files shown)


novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
```diff
@@ -153,12 +153,20 @@ final class SwapSetupPresenter: SwapBasePresenter {
             }
 
             payAmountInput = payAmount.map { .absolute($0) }
+
+            if quote.willCollectCommission {
+                receiveAmountInput = receiveChainAsset.map {
```
svojsu: I see there's a potential for cascaded decreasing of `receiveAmountInput` since the field is the source of truth for `refreshQuoteForBuy`. It reads it back into `receiveInPlank` and re-quotes from it. The setup interactor wires a debounced re-quote subscription:
 
`subscribeRequoteService` => `didReceiveQuoteDataChanged` => `refreshQuote` with preserved `.buy`

  The loop is:
  1. User types receive = 100.
  2. Quote arrives, presenter writes `receiveAmountInput` = 99.15 displaying post-fee value.
  3. If (or when) requote tick fires, `refreshQuoteForBuy` reads 99.15 and requests a buy quote for 99.15, handler overwrites `receiveAmountInput` = ~98.30 and repeats.

svojsu (changes requested):

Leemo94:
### 1. Swap Screen
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 54 15

This screen has been updated so that the service fee is reflected in the rate. The user cap tap the information button next to the rate to access the bottom sheet. Additionally, information about the fee is always available in this part of the UI.

### 2. Rate Bottom Sheet
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 54 20

### 3. Confirm Screen
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 54 30

Info about the rate is visible here too.

### 4. Swap Progress Screen 1
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 54 37

### 5. Swap Progress Screen 1
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 54 41

### 6. Txn details/history
Simulator Screenshot - iPhone 16 Prod - 2026-05-22 at 05 55 45

Leemo94:
Surfaced during sim QA: the transaction history still shows the Hydration Router system account (13UVJyLnzJK7QHxrrjLxwr58Lds4GR6Z37LxLtYrDm3qa5Je → modlrouterex derived account) as a separate transfer entry alongside the swap.

NovaFeeTransferFilter (TransactionHistoryTransfersFilter.swift:3) only filters the Nova fee account itself, not the router pallet account that  the omnipool uses as a middleman during trade execution. So the user sees the swap + two unrelated transfers (in and out of the router).

This pre-dates this PR — it's a Hydration-wide history quirk rather than a regression. Android caught the same thing during the parallel  2026-05-21 refactor and added a HydrationSystemAccounts.routerAccountId constant + extended the filter to cover both accounts (see RealHydrationSwapTransferFilterFactory.kt). Suggest mirroring that here: add a HydraConstants.hydraRouterAccountId constant (hex 6d6f646c726f7574657265780000...000000) and have the filter return false if either sender or receiver matches either the Nova fee account or the router account.

novawallet/Modules/Swaps/Setup/SwapSetupPresenter.swift
Leemo94: Done in `3a4a9016f`. The buy-direction branch of `handleNewQuote` no longer writes `displayAmountOut` back into `receiveAmountInput`; in buy direction it now only updates the pay field. `refreshQuoteForBuy` therefore always re-quotes from the amount the user actually entered, so the value can't be read back and decremented on each requote tick.

This brings it in line with the Android behaviour, where quote results never overwrite the field the user is editing. The post-fee receive amount is still surfaced through the rate, the rate info sheet, and the confirmation/execution screens (via `displayAmountOut`), so the fee stays visible.

Leemo94:
A note on the buy-direction display — flagging intentional behavior plus a possible follow-up, out of scope for this PR.

Since the fee is taken from the swap output, the two input directions surface it differently:
- **Sell** (user enters the amount to spend): the receive field shows the post-fee amount, so it matches what's received.
- **Buy** (user enters the amount to receive): the field keeps the entered figure, and the 0.85% fee is reflected in the rate and on the confirmation screen — so the user nets ~0.85% under the headline receive number.

This is deliberate and matches the Android PR (#2274), keeping the two apps consistent.

A possible future refinement (intentionally **not** in this PR): treat the buy-direction input as the desired *net* output and gross up the swap target, so the receive field equals what actually lands in the wallet. It would need to ship on both platforms together, and it'd be approximate rather than exact (the commission is computed from the post-slippage minimum). Happy to take that on as a separate coordinated change if wanted — otherwise the current behavior stands.

svojsu:
Implemented and released in 10.9.0

PR #1831, opened 2026-05-13
Staking: Bittensor (TAO) — root + subnet positions

Leemo94:
## Summary

First-class Subtensor staking on iOS for Bittensor mainnet. Users can stake to root (TAO) or any of the 128 subnets (alpha) via the standard Nova staking flow, with positions sourced from a single runtime API call and refreshed on every confirmed extrinsic.

**What ships:**
- Per-subnet rows on the multistaking dashboard — TAO root + one row per subnet where the user holds stake. Rows without stake are excluded from "Available to stake" so the dashboard doesn't list 128 empty options.
- Stake details screen rebuilt to match Nova's standard staking layout: your stake → Stake more / Unstake → inline "Your validator(s)" card → expandable "Staking info" footer. Scoped to the entry netuid.
- Full stake / unstake lifecycle against root or any subnet (`add_stake_limit` / `remove_stake_limit`). Slippage protection on the subnet path enforces a 0.5% cushion against the AMM spot price.
- Validator picker with on-chain-identity filter, "hide 100% commission" filter, and sort by total stake / APR. Subnet picker with sort by total TAO staked / subnet number.
- Smart-pick unstake UX: single position on the entry netuid → straight to setup; multi-position → action-sheet picker.
- Confirm flow pops back to the dashboard on `inBlock` and invalidates the position cache.

## Architecture

- **Position fetch** uses `state_call("StakeInfoRuntimeApi_get_stake_info_for_coldkey", coldkey)` — the runtime returns resolved `(hotkey, netuid, stake)` for every position, including root TAO and lazy-migrated subnet alpha. Storage-map walking was abandoned: `StakingHotkeys(coldkey)` does not always include root validators, and Bittensor lazy-migrated alpha shares from V1 (`Alpha` / `TotalHotkeyShares`) to V2 (`AlphaV2` / `TotalHotkeySharesV2`), so static readers silently miss positions. The runtime API is the canonical client path used by Talisman and SubWallet.
- **`SubtensorPositionCache`** (actor) is the single source of truth per coldkey. All 129 sync services share one in-flight fetch + 15s result cache. Cache invalidation uses a generation counter — it does **not** cancel the in-flight task, which previously propagated `CancellationError` to up to 128 awaiting services.
- **`SubtensorStakeInfoDecoder`** is the single SCALE decoder for `Vec<StakeInfo>`. Field order matches `pallets/subtensor/src/rpc_info/stake_info.rs`; if upstream inserts a field before `is_registered`, the skip count here is the only place to update.
- **Per-netuid consumer** `SubtensorMultistakingUpdateService` reads `typeExtras.netuid` from the ChainAsset, filters the cached result, writes one CoreData dashboard row, and re-polls on a 30s `syncUp(afterDelay:)` cadence.
- **Dashboard entry normalisation:** `StakingDashboardWireframe` rewrites the entry ChainAsset to native TAO (assetId 0) via `chainAsset.subtensorTaoAsset()` before opening the staking module, so units / balance / price always render in TAO regardless of which row was tapped. `entryNetuid` is threaded through.
- **Slippage encoding:** `add_stake_limit` / `remove_stake_limit` subnet branch encodes `limit_price` as `TaoBalance` (u64, RAO-per-Alpha) with `× 1_000_000_000.0` and `.rounded(.up)` on stake / `.rounded(.down)` on unstake. An earlier prototype multiplied by `2^32` (treating `limit_price` as I96F32 raw bits), which silently relaxed stake slippage by ~4.29× and would have rejected every subnet unstake outright. Three new tests in `SubtensorExtrinsicBuilderTests` cover subnet stake / unstake / missing-spot.

## Notes

- Companion chain config: novasamatech/nova-utils#4335 — adds 128 subnet-alpha assets (`assetId = 10000 + netuid`, `type = "subtensor-alpha"`, `typeExtras = {netuid: N}`, `staking = ["subtensor"]`, gold theme colour on the Bittensor chain).
- Android counterpart: novasamatech/nova-wallet-android#2283.
- Subnet names ship as placeholders (`SN1` … `SN128`). On-chain `SubnetIdentitiesV3(netuid)` has real names for a minority of subnets but most are unset and fetching all 128 hits public-RPC rate limits. Replacing the placeholders requires a runtime override layered on top — separate follow-up.
- Subnet-alpha icons reuse `TAO_bittensor.svg`. Per-subnet icons are a design follow-up.
- USD/fiat price for subnet alpha is not surfaced. Surfacing it requires a chain-local price provider reading `SubnetTAO / SubnetAlphaIn` (AMM spot) × TAO external price — separate follow-up.
- Public Finney RPC aggressively rate-limits and `state_getKeysPaged` returns `[]` there. The position cache is the chokepoint that makes 129 parallel sync services workable on the public endpoint.

### Indexer dependency

The `[TEMP-TAOSTATS]` scaffolding fills in everything Nova's Bittensor indexer is intended to provide. To remove the scaffolding cleanly, the indexer needs to expose:

- **Per-validator (hotkey) yield estimate** (APR / APY) — consumed by the validator picker rows and by the Start Staking Info "Earn up to %" headline. TaoStats publishes a 7-day rolling figure today.
- **Per-validator commission / take** — shown on each validator row and drives the "hide 100% commission" filter.
- **Per-validator total stake + nominator count** — sort key on the validator picker and row metadata.
- **Per-subnet aggregate APR** (root + each netuid) — dashboard tile headline yield per row. Without this the dashboard falls back to a placeholder 18% figure.


Leemo94:
**Note: single fee account for now; multiple accounts possible later**

This PR routes the fee to a single recipient (`SubtensorStakingConstants.novaFeeAccountId`) for all subnets, which is fine for v1. In future it may be worth supporting **multiple fee accounts** — e.g. one per subnet — if we ever want per-subnet fee attribution or finer-grained routing.

The code change would be small and contained: swap the single `novaFeeAccountId: AccountId?` for a lookup (`novaFeeAccountId(for netuid: UInt16) -> AccountId?`). The `netuid` is already threaded into `SubtensorCommissionFactory.make{Stake,Unstake}Commission` and the presenters' fee-applies check, so it's mostly plumbing.

The real open question is operational rather than technical — where the per-account addresses would come from: a hardcoded map (needs an app release per new subnet), a config-driven map (needs backend support), or a deterministic per-netuid sub-account derived from one key.

No action needed now — flagging it as a future option.

PR #1843, opened 2026-08-05
Claude system design workflow and harness

svojsu:


.claude/docs/process/design-loop.md
```diff
@@ -0,0 +1,384 @@
...
+| plan  | `plan-fidelity`    | `nova-plan-reviewer`  | opus / high    | nothing            |
+| plan  | `plan-correctness` | `nova-plan-reviewer`  | opus / high    | nothing            |
+| plan  | `plan-exec`        | `nova-plan-reviewer`  | opus / high    | nothing            |
+
+Reviewers are **read-only on the artefact**. Only `architect` writes `SPEC.md`; only `planner` writes
+`PLAN.md`. A reviewer that edits the document under review has destroyed the review.
+
+## Files
+
+Flat, one design in flight at a time.
```
stepanLav, line 45: **Suggestion: per-feature directories instead of a single fixed slot**

What about `.claude/designs/<feature-slug>/{SPEC,CONTRACTS,PLAN,REVIEW-LOG}.md` (with `designs/` gitignored and in `.git/info/exclude`) instead of five fixed paths? It would buy:

- **Parallel designs.** Right now a second feature cannot be specced while an approved-but-unexecuted plan exists — `/nova-spec` preflight has to make the human explicitly abandon it, because starting a new feature destroys the previous one's contracts.
- **History.** Clearing the slot erases the spec, the plan and the whole review record. When the shipped code has to be revisited months later, the reasoning is gone. Per-feature dirs keep it for free.
- **Simpler handoff.** A fair amount of machinery here exists specifically to defend against the slot being overwritten: the sha256 pinned in the plan header, the executor's pre-Task-1 hash check, the "a plan approved today can find different contracts under that path next week" caveat, the approved-unexecuted-plan guard in preflight. With per-feature dirs those become backstops instead of load-bearing mechanisms — the path itself is the identity.

The cost looks small: the skills take a slug (or keep an "active design" pointer), and the contamination story is unchanged — everything still lives under `.claude/`, so the `':(exclude).claude/'` pathspec and the guard assertion work as-is.

.claude/docs/process/review-lenses.md
```diff
@@ -0,0 +1,285 @@
...
+list the sections.
+
+**Zero findings is a valid outcome.** Say so. A long list of soft findings buries the one that
+mattered, and costs the author the attention that one deserved.
+
+---
+
+## Reference tables
+
+### Mandatory edge-case coverage (used by lens `reality`)
```
stepanLav, line 256: **Duplication: these reference tables (and the falsification block) exist in several copies**

- The mandatory edge-case matrix here is a verbatim copy of the one in `nova-architect.md` (Stage 3).
- The peer-file obligations table is duplicated in `nova-planner.md` ("Tasks that need particular care").
- The falsification shapes/test block appears in some form in five places: `design-loop.md`, this file, `nova-spec-reviewer.md`, `nova-plan-reviewer.md`, and `nova-cold-reader.md`.

These are exactly the tables that will get edited — a fifth staking flavour, a new failure shape — and the copies will drift. This PR itself fixes a dozen instances of that failure mode elsewhere in `.claude/docs/` (`EraValidatorsService` → `EraValidatorService`, `BaseDataValidatorFactory` → `BaseDataValidatingFactoryProtocol`, …), so we know how it ends.

Suggest one canonical home per table — this file's Reference tables section seems the natural one, since it is already deliberately process-free and the agents that need the criteria are already told to read it — with the agent definitions linking here instead of embedding a copy, the same way they already link to `design-loop.md` and the checklists. One extra `Read` per agent, and the next edit lands in one place.

stepanLav:
looks really good, added a few minor suggestions

stepanLav (approved):

PR #1845, opened 2026-08-09
Hydration swap fees

svojsu:
## Purpose

Charge a Nova Wallet fee of **0.85% of the amount received** on swaps that route through Hydration.
The fee is a transfer to a fixed Nova beneficiary account, batched into the **same atomic extrinsic**
as the swap, so the swap and the fee land together or not at all.

## How it works

**One number, computed once.** The commission is resolved from the quote and that exact value is used
for fee estimation, for the transfer that gets signed, and for every amount displayed. It is never
recomputed at submission. This is the load-bearing property — "displayed == charged" holds by
construction rather than by keeping several derivations in agreement.

```swift
struct AssetExchangeCommission {
    let chargingEdgeIndex: Int
    let asset: ChainAssetId
    let amount: Balance
    let beneficiary: AccountId
}
```

**Rate.** `AssetExchangeCommissionConstants.rate` is `85/10000`. 

**Where it is charged.** `AssetExchangeCommissionPolicy` is pure synchronous arithmetic with no
network reads. It selects the **last Hydra edge** in the path:

```swift
path.lastIndex { $0.type == .hydraSwap }
```

Consecutive Hydra edges merge into a single atomic operation, and because the charging edge is by
construction the last edge of its run, `AssetExchangeAtomicOperationArgs.extending(with:)` always
carries the commission into the merged operation. Routes with no Hydra leg get
`AssetExchangeNoCommissionPolicy`.

**Threading it through.** `AssetExchangeQuote` carries `commission` plus `metaOperationIndexByEdge`,
a real edge → meta-operation mapping built from the actual grouping, so display and attachment are
anchored to the same ground truth rather than to two rules that must agree.

**Buy direction.** `AssetsExchangeRouteManager` picks the winning path, then re-quotes it grossed up
by the rate so "buy exactly X" still delivers X after the fee. The gross-up is skipped when the path
has no charging site.

## What the user sees

- **Setup, Confirm, Execution and Route Details** show the receive amount **net** of the fee.
- The **rate** is computed from the net amount, so it already reflects the fee.
- A caption — `Includes 0.85% Nova fee.` — appears on **Setup only**.
- **Price difference** stays on the gross quote, since it compares against market price.
- **Existential-deposit validation** runs on net amounts: intermediate ED per hop, and destination ED
  against `worstCaseNetAmountOut`.

## History

The swap row shows the **gross** pool output and the beneficiary transfer is filtered out of history
(`HydrationSwapHistoryFiltersProvider`), matching the Android client. `TransactionHistoryTransfersFilter`
widened from `isBalancesTransfer` to `isTransfer` so the filter covers ORML and native commission
transfers alike, and `ExtrinsicProcessor` matches Hydra/AssetHub swaps **before** the generic transfer
matchers, since the batch now contains a transfer call and would otherwise be recorded as one.


novawallet/Common/Services/AssetExchange/Commission/AssetExchangeCommissionBeneficiaryProvider.swift
```diff
@@ -0,0 +1,115 @@
...
+}
+
+extension AssetExchangeCommissionBeneficiaryProvider: AssetExchangeCommissionBeneficiaryProviding {
+    func fetchStateWrapper(for chainId: ChainModel.Id) -> CompoundOperationWrapper<CommissionBeneficiaryState> {
+        if let cached = cachedState(for: chainId) {
+            return .createWithResult(cached)
+        }
+
+        do {
+            let chainAsset = try utilityChainAsset(for: chainId)
```
antonijzelinskij: Is it valid here to use utility chain asset? I guess we have to check balance in the same asset we pay comission.

antonijzelinskij (approved):

PR #1853, opened 2026-08-18
Fix Hydration swap fees

svojsu:
## Problem

Swaps on Hydration paying the fee in a non-native asset failed with `Token: FundsUnavailable`.
Max-amount swaps hit it every time, and every failed attempt still burned the fee.

The cause was fee under-estimation. To express the native fee in the fee asset we asked the router
for the cheapest swap path. But the chain never swaps the fee — `pallet-transaction-multi-payment`
converts it with a ten minute EMA oracle price along the route registered in `Router.Routes`. A
cheapest-path search deliberately lands on whichever pool sits furthest from the market, which is
exactly what makes it "cheapest", so our estimate came out systematically low.

## Fix

`HydraFeeOraclePriceCalculator` mirrors what the runtime does:

- takes the route from `Router.Routes`, with the same ordered-pair lookup, inversion and
  single-omnipool-hop default when no route is registered
- prices each hop the way the runtime does — omnipool through the LRNA hub, stableswap through the
  pool's share token, xyk directly, aave 1:1, and gives up on lbp/hsm rather than approximating
- ages stale oracle entries toward the last traded price before using them
- finishes with the runtime's own fixed point steps

The EMA maths itself is **not** reimplemented: it comes from `hydra-math-swift` 0.6.0, which exposes
Hydration's own `hydra-dx-math`. Only the parts that live in the pallets rather than in that crate
are written here — route resolution, leg expansion, and the final conversion.

The ten minute smoothing factor is derived from the chain's block time rather than hardcoded.
`pallet-ema-oracle` stores it as a constant table, but rebases that whole table whenever
`MILLISECS_PER_BLOCK` changes — v51 moves Hydration from 6s to 2s blocks and every entry moves with
it. The derivation uses the pallet's own `10 * MINUTES` grouping, where `MINUTES = 60_000 /
MILLISECS_PER_BLOCK`, so the integer division lands in the same place the runtime puts it.

The fallback order matches the chain: oracle price, then `AcceptedCurrencies`. There is no third
tier, because the chain has none — a swap quote is an execution price and is not equivalent. The
`AcceptedCurrencies` price is materially lower than the oracle price for most assets, so silently
landing there would reintroduce the same class of bug.

For an asset outside `AcceptedCurrencies` the chain does not use an oracle at all — it prices the fee
in DOT, buys that DOT on an xyk pool, and charges in DOT. We refuse to quote rather than return a
plausible wrong number.

Every storage read is pinned to one block hash. Otherwise oracle entries can arrive from a later
block than the block number they are aged against, which silently skips the ageing step.

`HydrationConversionFeePayment`'s equivalent — fees for ordinary extrinsics, transfers and XCM out of
Hydration — goes through the same path, so both call sites are fixed.

## Other changes

- `AssetExchangeFeeEstimatingRouter` now sends Hydration's `orml` and `ormlHydrationEvm` assets to the
  conversion estimator instead of the swap graph. AssetHub's `statemine` assets still use the graph,
  because AssetHub genuinely swaps the fee asset where Hydration only converts it.
- `HydraRouter.PoolType` gains the missing `HSM` variant. Reading `Router.Routes` widens the decode
  surface from routes Nova itself built to whatever governance registered, so all six runtime
  variants have to decode. Without it an HSM route would hard-fail fee estimation instead of falling
  back.
- `HydraFlowState`, `HydraQuoteFactory`, `HydraRoutesOperationFactory` and `HydraRoutesResolver` are
  deleted — they existed only to serve the old swap-quote fee path.
- `BlockHashOperationFactory` gains `createBestBlockHashWrapper` so the whole estimate can pin itself
  to one block.


novawallet/Common/Substrate/Types/HydraDx/HydraEmaOracle.swift
```diff
@@ -88,6 +88,22 @@ enum HydraEmaOracle {
...
+
+            return remainder > denominator / 2 ? quotient + 1 : quotient
+        }
+
+        static func tenMinutes(blockTimeMillis: BlockTime) -> BigUInt? {
+            guard blockTimeMillis > 0 else {
+                return nil
+            }
+
+            return value(forPeriodInBlocks: BlockNumber(tenMinutesDurationMillis / blockTimeMillis))
```
antonijzelinskij: As I see the final estimation is a bit another on chain:
OraclePeriod::TenMinutes => 10 * MINUTES as u64
pub const MINUTES: BlockNumber = 60_000 / (MILLISECS_PER_BLOCK as BlockNumber);

[traits/src/oracle.rs#L87-L96](https://github.com/galacticcouncil/hydration-node/blob/master/traits/src/oracle.rs#L87-L96)
[primitives/src/constants.rs#L55](https://github.com/galacticcouncil/hydration-node/blob/master/primitives/src/constants.rs#L55)

antonijzelinskij (approved):

novawallet/Common/Substrate/Types/HydraDx/HydraEmaOracle.swift
svojsu: fixed

